PR #80, opened 2014-05-15
[WIP] Split binary versioning and module finding logic into separate modules

springmeyer:
This work in progress starts splitting apart node-pre-gyp functions. It also adds a much cleaner method of requiring the binary module using [the latest node-bindings](https://github.com/TooTallNate/node-bindings/pull/12).

This will allow `node-pre-gyp` to only be a build dependency and not a runtime dependency. Usecase: avoid bloat. It will be safe to do `find ./node_modules -name node-pre-gyp | xargs rm -rf;` after `npm install` so that packaged applications like TileMill can avoid needing to ship dozens of node-pre-gyp's (and all its bundled deps) in its tree.

Mergable once:
- [x] - modules are fully split out and npm published
- [ ] - update `node-addon-example` to use new features


addaleax:
@springmeyer Anything one can do to help with this/one of the issues around this?

springmeyer:
This was a good idea, but is now old and stale. I don't have time to pick this back up myself, so closing.

Issue #106, opened 2014-08-26
free public hosting of binaries anyone ?


albanm:
Maybe this question doesn't belong to the scope of this project, but it is related.

I come from this well known npm issue https://github.com/npm/npm/issues/1891

It was closed as "node-pre-gyp has reached enough feature-completeness".

But it seems to me that node-pre-gyp while looking very good only responds to a part of the original need. What about hosting in the npm registry, is this forgotten ?

Honestly I don't intend to pay for an AWS S3 account to publish a small open source module. And anyway it doesn't seem a very sustainable approach to require every individual maintainers to do this.

So, is there any hope ?

maboiteaspam:
hi,

same problem here, i want to use node pre gyp, but can t afford the s3 amazon.
I believe that if many node users would be in, we could use bittorrent as storage backend.
feross provide full implementation in pure javascript.

if such network exists, would you be in ?

springmeyer:
@maboiteaspam  - node-pre-gyp does not have a hard dependency on AWS S3. It uses pure https to fetch binaries, which can be anywhere (and is configurable via the `host` option in package.json). So my preference would be to have anyone wishing to host elsewhere than s3 just write external scripts that call `node-pre-gyp reveal` in order to get the right info to publish binaries. More details on this at https://github.com/mapbox/node-pre-gyp/issues/134#issuecomment-69281215.

indieisaconcept:
Has anyone considered https://bintray.com/?

maboiteaspam:
Cool!
Hope licence and so on would let us push there the binaries needed by our respective github public projects.

PR #108, opened 2014-08-29
Pluggable hosting

albanm:
cf https://github.com/mapbox/node-pre-gyp/issues/106

Companion project: https://github.com/albanm/node-pre-gyp-hosting-github

The two together are tested in my fork of your node-addon-example project https://github.com/albanm/node-addon-example

I couldn't test that I didn't break s3 support as I don't have an account. I guess travis-ci will tell me soon enough if I did ok.

I didn't externalize s3 hosting in order not to break compatibility, but if you chose to do so it will be easy.

Of course I am open to all feedback.


indieisaconcept:
Is there any intention to continue with this PR? I was about to tackle something similar.

springmeyer:
@indieisaconcept - I would be interested in supporting tools outside of node-pre-gyp to automate the upload of binaries to custom hosts. This has the advantage of keeping node-pre-gyp simple, but not holding anyone back who thinks of yet another place they would like to store binaries. So, my proposal for how to do this is to write a script that hooks into node-pre-gyp either via command line calls (or programmatically). So you if want to push easily to bintray you could write a `pre-gyp-publish-binray` command or the like. More details on this at https://github.com/mapbox/node-pre-gyp/issues/138#issuecomment-74610853

indieisaconcept:
@springmeyer thanks. I was considering  revising node-pre-gyp to support the notion of publish adapters essentially removing all publish logic other than just simple orchastration.




I was then going to write separate adapters for s3, releases & bintray.




This PR looked like it was going in this direction hence the comment.




I think there is value in pulling s3 out of core though and I like the idea of a standardised way of describing a publish adapter. ( See travis/travis-dpl )




I would like to see adapters having an interface as such.




- publish

- unpublish

- info

albanm:
Hello. One thing this PR does not is: "holding anyone back who thinks of yet another place they would like to store binaries". Quite the contrary actually.

It makes node-pre-gyp pluggable, if the upload logic is already available in a module 'node-bintray' for example, just create a 'node-pre-gyp-hosting-bintray' module that bridges the gap.

If the upload logic is purely specific to a user, just write a 'node-pre-gyp-hosting-custom' plugin with a script loading mechanism and which would accept some parameters similar to this :
     
    "hosting": {
          "provider": "custom",
          "require": "my-script" // Path to the custom user's script
     }

The custom user script would have to implement the 'publish', 'unpublish' and 'download' methods just as other plugins.

Of course you don't have to create a separate 'node-pre-gyp-hosting-custom' project, just bring it with the main project as a officially supported hosting implementation in 'lib/util/hosting-custom.js' alongside with 'lib/util/hosting-s3.js'.

Hope this helps !

bchr02:
+1 for adding this directly in node-pre-gyp. It seems like it's a basic necessity. I almost didn't notice that the work was done and therefore almost wrote something similar. But more importantly the reason it should be included is because I consider storing binaries in github releases more of the default then S3. If there was one extra api that should be directly integrated into node-pre-gyp this is it.

piranna:
I agree with @bchr02, publish to GitHub is not recommended but it's free, while at the same time S3 is a paid option. As the README says GitHub is just enough for smaill projects, so I think this pull-request should be included and GitHub publishing set as default.

bchr02:
@piranna So you know, I ended up creating a module that complements node-pre-gyp called [node-pre-gyp-github](https://github.com/bchr02/node-pre-gyp-github) which allows publishing to GitHub releases.

Instead of ```node-pre-gyp publish``` you would use ```node-pre-gyp-github publish```

@springmeyer if your not going to add this support, maybe you can update your README.md to mention my module? Thanks.

springmeyer:
Closing since `node-pre-gyp-github` is the recommended solution and that is now documented at https://github.com/mapbox/node-pre-gyp#host

PR #148, opened 2015-02-17
Fix iojs win

springmeyer:
Makes adjustments to the tests to be able to pass with io.js + other refactoring that improves coverage and test consistency.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,7 @@ environment:
   matrix:
     - nodejs_version: 0.10.36
     - nodejs_version: 0.12.0
-    # not yet working (Module did not self-register), why?
-    #- nodejs_version: 1.2.0
+    - nodejs_version: 1.2.0
 
 platform:
   - x64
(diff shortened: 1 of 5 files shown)


Mithgol:
Somehow the changes made in mapbox/node-sqlite3@cdd8d450692224ea83f93264da8053ba251ebdee (they declare 2d72392dc8beff1c59770d9d5c4e1c3f6bf8a271 as the dependency) cannot prevent the message `Module did not self-register` on AppVeyor:

* [platform=x86](https://ci.appveyor.com/project/Mapbox/node-sqlite3/build/1.0.288/job/tpsne7l38yqa7jhf), line 624

* [platform=x64](https://ci.appveyor.com/project/Mapbox/node-sqlite3/build/1.0.288/job/bb65c4fnf9il30w5), line 654

springmeyer:
Those node-pre-gyp fixes only fixed the node-pre-gyp tests - they will have no impact elsewhere. But thanks for trying. So we must have some other problem in node-sqlite3?

Mithgol:
Ah, I've just misunderstood the scope of #148.

If only the tests are fixed, then we should still wait for a fix either in the engine (iojs/io.js#751) or in the building tool (TooTallNate/node-gyp#564); only then the generated modules won't cause that error message.

springmeyer:
Right, or we find a way to call `/path/to/iojs` directly. But I don't currently have good ideas to do that.

PR #151, opened 2015-03-08
Try to use npm_config_node_gyp env first

fengmk2:
npm@2.7.0 support custom node-gyp path,
we can use pangyp instead of node-gyp now.

https://github.com/npm/npm/pull/4887


techniq:
:+1:

springmeyer:
Thanks!

fengmk2:
:)

PR #157, opened 2015-04-29
Use semver@4.3.2

yhahn:
Per https://nodesecurity.io/advisories/semver_redos. The use of semver in node-pre-gyp is not exposed to this vulnerability but should reduce noise for people using retire/other security checking tools.


springmeyer:
merging, travis is failing for unrelated reasons: https://github.com/mapbox/node-pre-gyp/issues/158

Issue #162, opened 2015-05-05
node-pre-gyp as non bundled dependency


3y3:
How I understand the main reason to bundle `node-pre-gyp` is a usage it on installation phase.
And what happens if it will be used in `postinstall` step, is this resolves problem with bundling?
```
    "scripts": {
        "postinstall": "node-pre-gyp install --fallback-to-build",
    },
```

Unfortunately I can't test this now, but maybe the answer is on the surface?

springmeyer:
Moving to `postinstall` might work to avoid the chicken 'n egg problem of `node-pre-gyp` needing to be available at `install`. But `install` would still need to be overridden to prevent `npm` from detecting the `binding.gyp` and calling `node-gyp` to build. So, overall: maybe this would work but it would need a lot of testing and we'd need to figure out what would make sense to put into the `install` to make it do nothing.

3y3:
We need to redefine `npm preinstall` script. [proof link](https://docs.npmjs.com/files/package.json#default-values)

3y3:
@springmeyer , this works like a charm on my Windows machine:
```json
  "scripts": {
    "preinstall": "rem",
    "postinstall": "node-pre-gyp install --fallback-to-build"
  }
```

Can you test something like this for Linux or Mac. (I'm not sure which dummy command is preferable in bash).

rmg:
The `install` script is run after dependencies are installed. It is essentially a pre-postinstall, but with a default that auto-runs node-gyp.

3y3:
@rmg , my current solution (it tested on `v8-debug` and `v8-profiler`) is:
```bash
  "scripts": {
    "preinstall": " ",
    "postinstall": "node-pre-gyp install --fallback-to-build"
  }
```
Can you propose something more elegant?

rmg:
@3y3 nothing more elegant, sorry.

My point was just that you can use `install` instead of `postinstall`:
```json
 "scripts": {
    "preinstall": " ",
    "install": "node-pre-gyp install --fallback-to-build"
  }
```

springmeyer:
> My point was just that you can use install instead of postinstall:

[...]

PR #163, opened 2015-05-07
Update README.md: non bundled dependency tutorial

3y3:


bminer:
Why is this not the default approach?  Why is using `bundledDependencies` the recommended configuration?  Seems silly... and a waste of disk space.  node-pre-gyp already weights a whopping 4M, and telling package maintainers to bundle it instead of allowing npm to install it and manage dependencies seems just plain wrong.  Thoughts?

bminer:
+1 for merging

3y3:
There is opened issue where we discuss about this approach https://github.com/mapbox/node-pre-gyp/issues/162

springmeyer:
difference approach landed in https://github.com/mapbox/node-pre-gyp/pull/201

PR #174, opened 2015-09-23
Update README.md

etiktin:
Changes:
- Removed the "even Github recommends using it instead of github" remark about preferring S3, because the link doesn't state that (at least not any more). According to the page GitHub doesn't limit the size of binary releases or bandwidth used to deliver them. The only limit is that each individual file must be under 2GB in size. If that's the case maybe we should think about supporting GitHub as first class citizen.
- Capitalized "Travis" and "C++" for consistency.
- Removed the accidental remark at the end of the "Appveyor Automation" that pointed to Travis documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ### Features
 
- - A command line tool called `node-pre-gyp` that can install your package's c++ module from a binary.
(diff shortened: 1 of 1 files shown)


springmeyer:
Thanks for these improvements!

PR #175, opened 2015-09-24
Recognize electron as a valid runtime

zcbenz:
This doesn't make node-pre-gyp fully support Electron, but this makes `node-pre-gyp install --fallback-to-build` work for Electron when `npm_config_runtime` is specified, so modules relying on node-pre-gyp can be compiled against Electron by building from source.

I'll try to add full support for Electron in other PRs.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -17,6 +17,20 @@ if (process.env.NODE_PRE_GYP_ABI_CROSSWALK) {
     abi_crosswalk = require('./abi_crosswalk.json');
 }
 
+function get_electron_abi(runtime, target_version) {
+    if (!runtime) {
+        throw new Error("get_electron_abi requires valid runtime arg");
+    }
+    if (typeof target_version === 'undefined') {
+        // erroneous CLI call
+        throw new Error("Empty target version is not supported if electron is the target.");
+    }
+    // Electron guarentees that patch version update won't break native modules.
+    var sem_ver = semver.parse(target_version);
+    return runtime + '-v' + sem_ver.major + '.' + sem_ver.minor;
+}
+module.exports.get_electron_abi = get_electron_abi;
+
 function get_node_webkit_abi(runtime, target_version) {
     if (!runtime) {
         throw new Error("get_node_webkit_abi requires valid runtime arg");
@@ -55,6 +69,8 @@ function get_runtime_abi(runtime, target_version) {
     }
     if (runtime === 'node-webkit') {
         return get_node_webkit_abi(runtime, target_version || process.versions['node-webkit']);
+    } else if (runtime === 'electron') {
+        return get_electron_abi(runtime, target_version || process.versions.electron);
     } else {
         if (runtime != 'node') {
             throw new Error("Unknown Runtime: '" + runtime + "'");
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ addons:
   apt:
     sources: [ 'ubuntu-toolchain-r-test','llvm-toolchain-precise-3.5' ]
     # for testing node-webkit
-    packages: ['clang-3.5', 'xvfb','libasound2','libx11-6','libglib2.0-0','libgtk2.0-0','libatk1.0-0','libgdk-pixbuf2.0-0','libcairo2','libfreetype6','libfontconfig1','libxcomposite1','libasound2','libxdamage1','libxext6','libxfixes3','libnss3','libnspr4','libgconf-2-4','libexpat1','libdbus-1-3','libudev0']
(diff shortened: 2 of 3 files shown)


paulcbetts:
bless u @zcbenz, bless u

sheerun:
I'd love if you released it :)

We have issues with building electron project because fsevents 1.0.0 depends on node-pre-gyp.

paulcbetts:
@sheerun In the short term, `npm shrinkwrap` your project then hand-edit the shrinkwrap file to force fsevents to a lower version

springmeyer:
Thank you for this @zcbenz - will merge and release now. Therefore node-pre-gyp `v0.6.12` will support the `--runtime=electron` flag.

springmeyer:
`v0.6.12` is now published.

johnhaley81:
Thank you guys for this. <3

capelio:
Bless you all. :fireworks: :fireworks: :fireworks:

sheerun:
Yesterday I made a script to fix this issue. To feel better I'm pasting it here :)

```bash
#!/usr/bin/env bash

for path in $(find . -name package.json -type f -follow | xargs grep '"fsevents":' -l | sort | uniq); do
  (
    cd "$(dirname "$path")"
    VERSION=$(node -e "console.log(require('./node_modules/fsevents/package').version)" 2> /dev/null)

    if [[ "$VERSION" != "0.3.8" ]]; then
      echo "FIXING: $(dirname "$path")"
      npm uninstall --save fsevents
      npm uninstall --save-optional fsevents
      npm install --save fsevents@0.3.8
    fi
  )
done
```

suda:
If someone needs Electron detection (like Atom package which requires `node-pre-gyp`) I patched versioning.js here: https://github.com/suda/node-pre-gyp/tree/support-electron

kenr:
@zcbenz: One project I use (node-sqlite3) require electron set as runtime when importing AddOn  (https://github.com/mapbox/node-sqlite3/blob/master/lib/sqlite3.js#L3). I've created a PR that adresses this issue (https://github.com/mapbox/node-pre-gyp/pull/177).

lib/util/versioning.js
```diff
@@ -55,6 +69,8 @@ function get_runtime_abi(runtime, target_version) {
     }
     if (runtime === 'node-webkit') {
         return get_node_webkit_abi(runtime, target_version || process.versions['node-webkit']);
+    } else if (runtime === 'electron') {
+        return get_electron_abi(runtime, target_version || process.versions.electron);
```
cdaringe: what's the BKM for getting my runtime & target_version in, if I have a project that uses a package which consumes node-pre-gyp?

cdaringe: sorry, best-known-method. preferred method, really*

paulcbetts: So, the policy that https://github.com/electronjs/electron-rebuild uses is, assume your package is a sibling package of `electron-prebuilt`, then use https://github.com/electronjs/electron-rebuild/blob/master/src/main.js#L42 to get the version info. 

This isn't perfect, but it makes the Easy case Easy, and if you let devs also specify explicitly, you have an escape hatch

cdaringe: hmm. so if I'm on `electron-rebuild@1.0.1` and running it on `postinstall`, it should set it for me, eh?  curious why I still get the `Error: Empty target version is not supported if electron...` error from above

paulcbetts: It will if `electron-prebuilt` is in the same folder as `electron-rebuild` in your `node_modules` folder. If not, nbd, you can use the cmd line flags to work around it

cdaringe: thanks @paulcbetts, really appreciate it.  perhaps you'd be willing to help me ID what I've missed.  electron-prebuilt and electron-rebuild are good to go, it would seem.
```
cdieringer@Snapper-osx:~/node/coinstac$ ll node_modules/ | grep electron
drwxr-xr-x  15 cdieringer  staff   510 Oct 12 13:36 electron-prebuilt/
drwxr-xr-x  10 cdieringer  staff   340 Oct 12 14:03 electron-rebuild/
```
```
cdieringer@Snapper-osx:~/node/coinstac$ cat package.json | grep electron
    "electron-prebuilt": "^0.33.7",
    "electron-rebuild": "^1.0.1",
```

however, it still yields the following, against expectation:
```
...
> fsevents@1.0.2 install /Users/cdieringer/node/coinstac/node_modules/babel/node_modules/chokidar/node_modules/fsevents
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! install error
node-pre-gyp ERR! stack Error: Empty target version is not supported if electron is the target.
```

paulcbetts: Oh this is a different error than I thought it was - I'll look into this. Can you file a bug on electronjs/electron-rebuild?

cdaringe: ya, id be happy to. i might spend a little time looking into it tonight.  i'd really like to give back somehow to a project that has helped me so much!

PR #177, opened 2015-10-06
Fixing issue with electron support when importing AddOn/binding 

kenr:
This PR fixes an issue when importing AddOn/binding when runtime is electron.

PR #187, opened 2015-11-05
Detect the correct runtime when find() is called from Electron

enlight:
This fixes the case where require('node-pre-gyp').find() is called from Electron without explicitly specifying the runtime in the options. Previously the runtime would just default to "node" and find() would fail to find any native modules specifically built for Electron because `node-gyp` named the build output directory `{runtime}-v{version}-{platform}-{arch}`. For example after building a native module for Electron v0.34.2 the module binary ended up in the `electron-v0.34-win32-x64` directory, but find() was looking for it in the `node-v46-win32-x64` directory.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -241,6 +241,16 @@ function drop_double_slashes(pathname) {
     return pathname.replace(/\/\//g,'/');
 }
 
+function get_process_runtime() {
+    var runtime = 'node';
+    if (process.versions['node-webkit']) {
+        runtime = 'node-webkit';
+    } else if (process.versions.electron) {
+        runtime = 'electron';
+    }
+    return runtime;
+}
+
 var default_package_name = '{module_name}-v{version}-{node_abi}-{platform}-{arch}.tar.gz';
 var default_remote_path = '';
 
@@ -249,7 +259,7 @@ module.exports.evaluate = function(package_json,options) {
     validate_config(package_json);
     var v = package_json.version;
     var module_version = semver.parse(v);
-    var runtime = options.runtime || (process.versions['node-webkit'] ? 'node-webkit' : 'node');
+    var runtime = options.runtime || get_process_runtime();
     var opts = {
         name: package_json.name,
         configuration: Boolean(options.debug) ? 'Debug' : 'Release',


enlight:
There's actually an existing PR https://github.com/mapbox/node-pre-gyp/pull/177 to address this issue, unfortunately I only noticed it after implementing my own fix. Compared to the previous PR mine is a single commit and more conformant to the existing code style. Anyway, I'm not fussed about which one gets merged in as long as the issue is fixed.

Ajusa:
Can someone merge this? I need this for my latest project...

maxkorp:
1) These test failures seem to strictly be CI hiccups with network connectivity or missing fixtures, not problems with the code/modification.

2) I made an additional module to sort of support this, until things start getting merged at least. It doesnt look at process versions but instead at config from an npmrc, although process.versions could be easily added as well. 

https://github.com/maxkorp/node-pre-gyp-npmrc-helper

Ideally, I'd like to get the npm config lookup and 187 added to just node-pre-gyp.

lukeapage:
I can confirm this branch works. Please merge!

lukeapage:
@springmeyer any chance you can take a look ?

enlight:
To anyone else who stumbles in here I've created [electron-node-pre-gyp-fix](https://github.com/enlight/electron-node-pre-gyp-fix) to monkey-patch `node-pre-gyp` at runtime, some may find it preferable to relying on the branch in this PR.

enlight:
Actually, the `node-pre-gyp` workaround implemented in [electron-rebuild](https://github.com/electronjs/electron-rebuild) is more robust than my monkey-patch, the monkey-patch doesn't seem to work on `node-inspector`.

maxkorp:
For what its worth, with the current version, just using an npmrc with the appropriate target and versions set will suffice for npm install to download the appropriate binaries, or compile appropriately when needed.

barbalex:
please merge this

springmeyer:
> Actually, the node-pre-gyp workaround implemented in electron-rebuild is more robust than my monkey-patch, the monkey-patch doesn't seem to work on node-inspector.

This patch looks good and mergable but I held back due to the above comment. @enlight - can you share more details on whether you think this should land or not?

lukeapage:
@springmeyer that comment is specific to electron-rebuild and what they are doing - i dont think it is a criticism of this patch. Currently electrons primary instructions for how to debug include installing a patched version of node-pre-gyp - http://electron.atom.io/docs/v0.37.3/tutorial/debugging-main-process/

enlight:
@lukeapage is correct @springmeyer my earlier comment was in relation to one of the workarounds people have had to employ while waiting for this PR to get merged, once this PR is merged those workarounds will no longer be necessary.

springmeyer:
Thanks @enlight - I'll merge and tag `v0.6.24` now with this fix. Thanks!

barbalex:
Yuppi aye, aye, yuppi yuppi aye...

Thanks a lot to all!

hedefalk:
This did not work for apm, but looking at this commit I tested just setting --runtime=electron and all my troubles was gone. Made a PR for apm: https://github.com/atom/apm/pull/536

Steviey:
module.js:442 Uncaught Error: Cannot find module
...sqlite3\lib\binding\electron-v1.3-win32-x64\node_sqlite3.node'

chrisveness:
Hmm... have I hit the same (or similar) problem?

    Uncaught Exception: Error: Cannot find module '.../app/node_modules/sqlite3/lib/binding/electron-v1.4-linux-x64/node_sqlite3.node'
      at Module._resolveFilename (module.js:455:15)

Noob on Electron trying to find my way around, not sure where node-pre-gyp fits in, hitting problems with sqlite3...

Using Electron 'two package.json', Electron v1.4.1, node-pre-gyp v0.6.28 from sqlite3 v3.1.4, node v6.6.0, npm v3.10.3 on Linux.

Any hints what do to?

PR #193, opened 2016-01-04
GitHub suggests using GitHub Releases now

rgbkrk:
This link no longer suggests relying on S3, likely because GitHub Releases supports raw binaries and is backed by S3 now (as is Large File Support).


springmeyer:
Thanks @rgbkrk

PR #199, opened 2016-03-10
Simple tweak to support building in Electron

giseburt:
I know there are other pull requests that are similar. This one I have tested and works without other hand holding. It builds on other recent work to support electron, but auto-detects electron from the process.versions.

_Update:_ I see it's failing checks. I'll look at those as soon as possible.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -250,9 +250,9 @@ module.exports.evaluate = function(package_json,options) {
     var v = package_json.version;
     var module_version = semver.parse(v);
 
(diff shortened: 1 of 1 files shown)


giseburt:
Well, now it passes on everything but node 5.

giseburt:
The failures don't appear to be related to this pull request.

springmeyer:
landed the equivalent fix in https://github.com/mapbox/node-pre-gyp/blob/master/CHANGELOG.md#0625

PR #206, opened 2016-04-22
replaced .find reliance on deprecated fs.existsSync

bchr02:
```javascript
var existsSync = require('fs').existsSync || require('path').existsSync;
```

I can't find existsSync in path. I looked as far back as the 0.10.x docs. Also, existsSync is deprecated.

I replace it with fs.statSync which has been around at least since 0.10.x probably longer.


bchr02:
I had to recreate my fork, so I am closing this PR. I have reopened it [here](https://github.com/mapbox/node-pre-gyp/pull/217)

PR #217, opened 2016-06-07
replaces find fn's reliance on deprecated fs.existsSync

bchr02:
I can't find existsSync in path. I looked as far back as the 0.10.x docs. Also, existsSync is deprecated.

Therefore, I replaced it with fs.statSync which has been around at least since 0.10.x probably longer.

### DIFF
--- a/lib/pre-binding.js
+++ b/lib/pre-binding.js
@@ -1,8 +1,17 @@
 "use strict";
 
 var versioning = require('../lib/util/versioning.js');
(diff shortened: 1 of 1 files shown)


springmeyer:
Closing without merging `Note that fs.exists() is deprecated, but fs.existsSync() is not.` at https://nodejs.org/api/fs.html#fs_fs_existssync_path does not indicate `fs.existsSync` is deprecated.

Issue #209, opened 2016-05-03
node-pre-gyp Appveyor build failing for Windows x86 using Node 4.4.3


deadprogram:
I was having trouble building for Windows x86 using Node.js 4.4.3, when I noticed that the Appveyor build for node-pre-gyp itself is failing for the same reason:

```
ENOENT: no such file or directory in readCAFile
```

The failed build can be seen here:

https://ci.appveyor.com/project/Mapbox/node-pre-gyp/build/1.0.373/job/h12w3ymhbyg6ltq6#L481

The build for Windows x64 using Node 4.4.3 succeeds, however.

I see that the node-gyp command is failing, but I thought perhaps node-pre-gyp might be missing something for the setup for the 4.4.3 release.

Any thoughts as to what is missing here? Now that 4.4.3 is the LTS version, it would be great to be able to have nice precompiled binaries for it.

Thank you!

springmeyer:
This looks like something broken in node-gyp. node-gyp is failing in readCAFIle which recently was added in https://github.com/nodejs/node-gyp/pull/837 /cc @bnoordhuis

bnoordhuis:
It looks like something (.npmrc? something in the environment?) is setting `npm_config_cafile` to the string `'undefined'`.

reconbot:
Should this be filed on node-gyp itself?

reconbot:
I've got a two step work around and I filed a bug with node-gyp.

The error we see is that the string `undefined` isn't a file. I'm only seeing this on 32bit windows machines with node 4.3 and 4.4. I haven't checked others.

The work around is to set the `cafile` option to a real file, I chose `package.json`. Then we get a second bug where even though it hasn't found a cert in package.json it still tries to use one. So you have to disable `strict-ssl `. We get some warnings but it works and doesn't seem to affect other node versions.

I added these to install phase in `appveyor.yml`
```yaml
  # work around an issue with node-gyp v3.3.1 and node 4x
  # package.json has no certificates in it so we're cool
  # https://github.com/nodejs/node-gyp/issues/921
  - cmd: npm config set -g cafile=package.json
  - cmd: npm config set -g strict-ssl=false
```

reconbot:
[...]

PR #243, opened 2016-10-23
Honor npm configuration for CA bundles

heikkipora:
Pass ‘ca’ and ‘cafile’ values from npm config options to request() when downloading binaries to honor a user-specified SSL CA. Allows using node-pre-gyp on environments with custom CAs.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -36,6 +36,16 @@ function download(uri,opts,callback) {
         }
     };
 
+    if (opts.cafile) {
+        try {
+            requestOpts.ca = fs.readFileSync(opts.cafile);
+        } catch (e) {
+            return callback(e);
+        }
+    } else if (opts.ca) {
+        requestOpts.ca = opts.ca;
+    }
+
     var proxyUrl = opts.proxy ||
                     process.env.http_proxy ||
                     process.env.HTTP_PROXY ||
@@ -159,6 +169,10 @@ function install(gyp, argv, callback) {
         } catch (err) {
             return callback(err);
         }
+
+        opts.ca = gyp.opts.ca;
+        opts.cafile = gyp.opts.cafile;
+
         var from = opts.hosted_tarball;
         var to = opts.module_path;
         var binary_module = path.join(to,opts.module_name + '.node');


heikkipora:
<sigh> https://github.com/RetireJS/retire.js/issues/141

heikkipora:
Retire.js also incorrectly flags cli-1.0.1 as vulnerable, so posted a PR for that as well to get your build to pass 😄   https://github.com/RetireJS/retire.js/pull/143

springmeyer:
Thank you for the contribution @heikkipora. I have a few questions:

1) This looks different than what node-gyp does (https://github.com/nodejs/node-gyp/commit/8c4b0ffa503aa385f4186e028542dcedca2efeee). I think the node-gyp impl looks more robust. Can you take a look at reimplementing this based on node-gyp's method? I think that `gyp.opts.cafile` will be == to `opts.cafile` in node-pre-gyp.

2) We'll need tests for this functionality. Can you think of a way to test this in the unit tests?

heikkipora:
1) I originally took a look at that impl and thought that readCAFile() is practically just waste as people don't generally store random stuff in their PEM files 😄    - and reading a PEM directly without parsing their content is common practice (see e.g. https://github.com/request/request#tlsssl-protocol).

All of gyp.opts.* are currently not merged to opts.* so opts.cafile won't be available automatically. I modified my commit so that gyp.opts.ca and gyp.opts.cafile are passed to download() via opts - which seems a bit cleaner approach to accessing them from process.env

2) I cannot figure out any meaningful way to test this with a unit test. On the other hand it's not very critical nor complex functionality so it may be OK to let it live without one (writing a complicated test setup just for this doesn't have a good ROI in my opinion).

What do you think ?

springmeyer:
Looks great @heikkipora, thank you for that revision. I notice that something is broken with appveyor and it appears to be in old tests that are no longer maintained. I just removed those in https://github.com/mapbox/node-pre-gyp/commit/35b10c32926c62817be6a6b91eaa5cadd45f3d5a. Would you mind rebasing against master and then we'll see if your commit passes on both travis and appveyor okay?

heikkipora:
The PR has now been rebased against master. However, there's still one (rather ironic) failure on the AppVeyor node 4.6.1/x86 environment where **node-gyp** tries to read a PEM CA file with undefined name 😃 .  Do you have any clue why it does that or how to make that work?

springmeyer:
@heikkipora - thanks for the rebase. About to head to sleep here (PST) but will plan to merge in the morning. As far as that bizarre appveyor test - no, mysterious (https://github.com/mapbox/node-pre-gyp/issues/209) and unrelated to your PR (though I hoped your PR might solve).

heikkipora:
Ok!

springmeyer:
Merged, will be included in v0.6.32. Thanks @heikkipora!

PR #252, opened 2016-11-08
Update dependency version to support nodejs till v0.12.x

manixate:
https://github.com/request/request/commit/01aefdd87cdd9cc44af6828c55fe5ef9c7e205d7 dropped support for nodejs 0.12.x so we should use till `v2.76.x`


springmeyer:
done in https://github.com/mapbox/node-pre-gyp/commit/479fe61942462fbc0fea7220b56d17871742d4b2 and available in the newly released node-pre-gyp `v0.6.32`

wtgtybhertgeghgtwtg:
I think [`479fe61`](https://github.com/mapbox/node-pre-gyp/commit/479fe61942462fbc0fea7220b56d17871742d4b2) actually does the opposite of this, since this was to lock `request` to `2.76.0`, and this bumps `request` to at least `2.79.0`.

manixate:
@wtgtybhertgeghgtwtg exactly. @springmeyer This pull request was meant to make `node-pre-gyp` to be compatible with nodejs v0.12.x as `request` version that is included doesn't support mentioned nodejs version.

springmeyer:
@wtgtybhertgeghgtwtg @manixate - as I see. Sorry for my confusion. However I'm not going to pin to an older request version. The reason is that node v0.10.x and v0.12.x are, respectively, end-of-life and end-of-life in less than a month per https://github.com/nodejs/LTS/#lts-schedule. So support for them is ending in the upcoming releases of node-pre-gyp.

wtgtybhertgeghgtwtg:
And I'm all for that.  Does this mean dropping `request` for `got` will be looked at again?

springmeyer:
> Does this mean dropping request for got will be looked at again?

Yes, I think we should revisit that. Although both the move to request-core and/or got would need more testing help. request-core work had to be reverted since publishing was failing and I've not had time to look into it.

Issue #253, opened 2016-11-18
HTTP Request fails when behind Proxy on Windows


FreeMasen:
I have experienced this in other npm packages that require('request').  I see you are looking for proxy env variables, however it is important to also to allow for  `requestOpts.strictSSL = false` and possible `requestOpts.tunnel = false` though I have not done enough testing to say if the latter makes any difference .

I am re-running an install attempt now, will update if setting this fixes the issue I am experiencing.

PR #254, opened 2016-11-19
update request to allow for strictSSL = false

FreeMasen:
By allowing the env variable strictSSL = false to be included in the install request, it makes it easier for users behind a proxy to use this application.  The default for Request is true if the variable is undefined in the Opts object, so the above change would allow a user to override the default if desired. Other env variable formats could be included.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -55,7 +55,7 @@ function download(uri,opts,callback) {
         log.verbose('download', 'using proxy url: "%s"', proxyUrl);
         requestOpts.proxy = proxyUrl;
         requestOpts.strictSSL = opts.strictSSL ||
-                    process.env.strictSSL
+                    process.env.strictSSL;
       } else {
         log.warn('download', 'ignoring invalid "proxy" config setting: "%s"', proxyUrl);
       }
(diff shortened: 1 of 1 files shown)


FreeMasen:
This is to address the problem in issue #253

springmeyer:
Thanks, however I'm a little hesitant to accept this because I've not encountered this issue myself. Can you link to other big users of request that have implemented this? Should the fix be ported to request itself? Is setting strictSSL off a potential security issue? Should this value come from `npm config` settings instead of the environment?

FreeMasen:
strictSSL = true is the default implementation at Request.  From behind a corporate proxy (why I am encountering these issues), if this is not set to false all Requests will fail. 

The two notable places where this setting is available would be NPM and VSCode. As far as whether or not they are using Request, I am not sure. 

It might be better to reference the npmrc variable `strict-ssl` 

While setting this false technically makes things less secure, it being by default set to true would indicate that any user leveraging this should have a reason. 

I can understand if you are unwilling to accept, it was a bit of a pain for me to try any use this so I thought I would try and make other's experience easier.

PR #259, opened 2016-12-15
Include package.json instead of just package

drewloomer:
This probably seems unnecessary, but tools like Rollup that do transforms based on file extension choke when trying to understand loading a JSON file without an extension specified. This is stopping me from using libraries which depend on node-pre-gyp in my build process.


tmcw:
👍 Thanks! Test failures are unrelated.

PR #261, opened 2017-01-12
Update README.md

GeorgeBgk:


springmeyer:
Thanks for you contribution. However node-pre-gyp does not depend on python when installing from binaries. To compile binaries it depends on node-gyp and that is when python is required, by node-gyp. I'll accept a PR that adds this detail.

Issue #101, opened 2014-07-28
cache builds locally.


bukzor:
@springmeyer and I discussed this at some length on freenode #mapbox last Friday.
Here are my notes: https://github.com/bukzor/node-pre-gyp/commit/c6481ae384a86bc9f3799fb666076b4d08584f6b

The general idea is that if you build something with node-pre-cache twice, the second build should be cached and very quick, even if the upstream manager has not published a build for your platform.

PR #272, opened 2017-03-15
Add support for caching downloaded bindings.

bruce-one:
This is some initial work on storing downloaded bindings in a `~/.node-pre-gyp` directory and reusing them if they're already present.

It will ignore/replace invalid cached bindings if the extraction process reports an error.

From what I can tell it works? (I seem to have issues with my env because the existing tests don't pass, even without the changes here?) (I tested the functionality on a simple project which did what I expected.)


springmeyer:
@bruce-one - thanks for the contribution!

> the existing tests don't pass, even without the changes here

Can you sync with latest master? This should be working now such that linux tests pass again.

bruce-one:
It looks like it's already up to date :-)

It actually seems to be something esoteric about my computer.... (Just double checked - all the `passes down --arg to node-gyp via X` seem to struggle for me; but they all pass on CI.)

The CI _mostly_ passed - looks like there was one fairly odd seeming issue on Travis (it seems to hang mid download?); and then just one issue on Windows which looks to be because an artefact (https://node-pre-gyp-tests.s3-us-west-1.amazonaws.com/app1-v0.1.0-node-v46-win32-ia32.tar.gz) is missing?

bruce-one:
(Just trying to retrigger the CI to see about that Linux test.)

springmeyer:
> It actually seems to be something esoteric about my computer.... (Just double checked - all the passes down --arg to node-gyp via X seem to struggle for me; but they all pass on CI.)

Sorry about the hassle. The tests could use some love for sure. One issue is that NODE_PATH and PATH are modified dynamically to ensure node-pre-gyp can use itself during testing. If your node setup depends on modifying these to work, that may be the conflict.

> The CI mostly passed - looks like there was one fairly odd seeming issue on Travis (it seems to hang mid download?); 

Yes, looks like a network blip (https://travis-ci.org/mapbox/node-pre-gyp/builds/211207496). Nothing on your end.

> just one issue on Windows which looks to be because an artefact (https://node-pre-gyp-tests.s3-us-west-1.amazonaws.com/app1-v0.1.0-node-v46-win32-ia32.tar.gz) is missing?

The appveyor build failure you can ignore (tracked at https://github.com/mapbox/node-pre-gyp/issues/209).



https://travis-ci.org/mapbox/node-pre-gyp/builds/211207496

bruce-one:
Added support for caching local builds as well (a la #101).

I'm not sure whether the names of the cache-related arguments are logical/sensible... Happy to change them if something else makes more sense :-) (Initially I was thinking "ignore cache" == "ignore the existing cached value"; and "skip cache" == "don't update the cache" (which were two different things in my mind (and might be used slightly differently) but not sure if that's just overthinking it :-p ))

springmeyer:
Thanks for working on this. I think those options are logic and named fine. I guess my one concern is that we try to model behavior based on how `npm` and `node-gyp` work, if possible. Do they support options like these to avoid cache usage? Or did you invent these yourself?

My understanding is that npm has `npm cache clean` to remove items. Does it have a way to skip using the cache without removing them? If not should `node-pre-gyp` also do this: instead of providing an option to skip should node-pre-gyp just expect that you manually have cleared your cache?

bruce-one:
(Tldr at the bottom, this got a bit waffly :-p )

Yeah, that's a fair question - and something I'd wondered about myself. (It's worth noting, I mostly just added those options because they seemed like something someone else might want, not something I'd expect to use myself.)

During making the PR I actually considered adding a `node-pre-gyp cache` subcommand (modeled after the npm or yarn ones)...

But... (For me) I actually don't use `node-pre-gyp` like I use npm or yarn. eg: Other than building packages I don't interact with the `node-pre-gyp` command directly (eg when installing a package that uses node-pre-gyp). Hence, how would I use that command? (doubly so if my node_modules is empty, but I don't want to use the cache, I can't run the command because node-pre-gyp isn't there yet, but node-pre-gyp will use the cache by default...) And/or, know that I want to? 

I went with the arguments modeled how they are because it meant that these flags could be used similar to how `npm install --fallback-to-source` (et al) work. eg `npm install --ignore-cache`, because maybe I know the upstream artefact changed, or something like that. Or `npm install --skip-cache` because disk space is important to me, and I don't want the node-pre-gyp cache clogging it up.

As I wrote that, though, I realised that it's a bit unclear that they're node-pre-gyp flags, rather than npm flags. If going argument style they'd probably deserve to be more clearly related to node-pre-gyp by name (eg `--skip-node-pre-gyp-cache` ish...).

Another thought was to use the `npm config` functionality for these flags -- eg, if you don't want to use the node-pre-gyp cache run `npm config set node-pre-gyp-cache false` (which has the effect of both flags).

My thought:
    - Add support for an npm config flag to disable the node-pre-gyp cache globally (`npm config set node-pre-gyp-cache false` vs `npm config set node-pre-gyp-cache true` (default to true))
    - Keep the arguments, but rename them to be more clearly node-pre-gyp related (`--ignore-node-pre-gyp-cache` and `--skip-node-pre-gyp-cache`?) Even with the subcommand, this have a use (imo)
    - Maybe/probably add a cache clean subcommand, eg `node-pre-gyp cache clean` (maybe with `ls` and `dir` as well?) (I actually think this is of lower value?)

lib/install.js
```diff
@@ -191,8 +191,9 @@ function install(gyp, argv, callback) {
         var to = opts.module_path;
         var binary_module = path.join(to,opts.module_name + '.node');
 
-        opts.ignore_cache = gyp.opts['ignore-node-pre-gyp-cache'];
-        opts.skip_cache = gyp.opts['skip-node-pre-gyp-cache'];
+        var no_caching = process.env.npm_config_node_pre_gyp_cache === 'false';
```
bruce-one: I'm tempted to call `npm config get node_pre_gyp_cache` (or something of that ilk; ignoring failure) if it doesn't look like it's running under npm, because otherwise `npm i` could be different to `node-pre-gyp install`, if that makes sense?

springmeyer:
@bruce-one - apologies I've not had a chance to digest this fully still. How close do you feel this is to being ready to merge. What additional feedback do you need?

bruce-one:
Good question :-p

It seems to be working fairly well for me...

I'm not really sure; it feels pretty ready to me?

tchakabam:
Hi, just wondering, what is the status here? I would be very interested by this feature, and possibly finishing it to be functional.

springmeyer:
@tchakabam thanks for sharing your interest. I think this is a desirable feature but also one that could create confusion/regressions unless robustly. So, I've personally not wanted to take that burden on (testing extensively). But someone (like you) motivated to make it work could definitely finish this. It might make sense to introduce the feature as opt-in only (default off) to avoid the chance of breaking users.

tchakabam:
I see, thanks for your quick reply. I am wondering: Instead of setting an HTTP download location, could it also be a location that is resolved from the package assets themselves (already downloaded by npm with package tar). 

Then we would be able however to take advantage of node-pre-gyp platform detection, and the naming-schemes configurable in the package-json to lookup actual files per platform/arch. 

But my aim would be in fact to not have to download anything from anywhere else than NPM in the first place :)

springmeyer:
@tchakabam node-pre-gyp exists for the usecase where it feels wrong (either due to a massive size, or ability to control outside of npm) to store inside the npm tarball. If you have that usecase then https://github.com/prebuild/prebuildify exists and works better for that.

tchakabam:
Ok, I see! Thanks for the pointer :)

tchakabam:
@springmeyer The way you put it, I understand better the motivation behind this project. 

Still, I think that if I would use this large-scale, I couldn't live with the idea there is no caching option here, just for the idea of unnecessary bandwidth/traffic used, or even unnecessary duplication of binaries instead of sym-linking to a single cache dir per user home.

tchakabam:
Another question: afaik npm for example uses md5 hashing on publish to ensure data integrity (and also avoid mitm exploits from modifying package content). So there isn't any such mechanism here, and that would be the problem in the first place with verifying integrity of cached data, as you mentioned somewhere?

> It might make sense to introduce the feature as opt-in only (default off) to avoid the chance of breaking users.

of course that makes sense, otherwise we would break the existing contract this tool exposes.

springmeyer:
> just for the idea of unnecessary bandwidth/traffic used, or even unnecessary duplication of binaries instead of sym-linking to a single cache dir per user home.

Yes, caching in a single directory would be a nice solution to that expense. However, at least for my usecases these days, my projects all get packaged up as docker images and that packaging happens on a remote machine with fast access to s3. So, I'm personally less offended by the unnecessary bloat between projects now than I was in the past.

> Another question: afaik npm for example uses md5 hashing on publish to ensure data integrity (and also avoid mitm exploits from modifying package content). So there isn't any such mechanism here, and that would be the problem in the first place with verifying integrity of cached data, as you mentioned somewhere?

Correct, node-pre-gyp lacks that support but its been requested and you raise a good point that it is likely a requirement of doing caching correctly.

springmeyer:
this got stale. if others want to pick this up, please create a fresh new PR. thanks.

PR #275, opened 2017-03-21
Node v4.8.1.

bruce-one:


springmeyer:
Thanks. I think instead of having to update these to keep the tests passing we need a better solution. I've sketched out an idea at #276. Would love help implementing if you are interested.

springmeyer:
Just committed 737d9cf to update the versions. Looks like your PR was failing since it missed some. Will close this now. Would love any feedback on #276

Issue #278, opened 2017-03-28
pre-gyp looks for the wrong ABI in electron childProcess.forks


bcomnes:
It appears node-pre-gyp modules are trying to use the wrong ABI when run inside of a childProcess.fork from an electron process, even if you have rebuilt the correct ABI using something like [electron-rebuild](https://github.com/electron/electron-rebuild).  It looks like this problem has been around a while, and even a [few](https://github.com/electron/electron-rebuild/commit/496439a7c20a8ce08ea3bc242a440187948dff53) [hacks](https://github.com/electron/electron-rebuild/blob/27fd3114b68bf5f9c438255b412cea763e768826/src/node-pre-gyp-fix.js) came and went.

Here is an example of the issue with sqlite3: https://github.com/ballpit/electron-sqlite3-fork-bug/blob/master/main.js

It looks like [node-gyp-build had a similar issue](https://github.com/electron/electron/issues/8727) a while back, and they ended up implementing a check for the [process.env.ELECTRON_RUN_AS_NODE](https://github.com/mafintosh/node-gyp-build/blob/master/index.js#L51) and returning electron bindings when thats found, and it looks like maybe we need to port that over to node-pre-gyp as well to fix this problem.  

It looks like the ABI selection is happening in https://github.com/mapbox/node-pre-gyp/blob/master/lib/util/versioning.js. Any idea where such a check might live in that file?

Tomyail:
any update for this issue? I need this pull request

Tomyail:
I create a [pull request](https://github.com/mapbox/node-pre-gyp/pull/343) for this issue

bcomnes:
This is my original PR to close this: https://github.com/mapbox/node-pre-gyp/pull/279

Open to either solution, but I think we should aim to make this work without user provided env vars.

bcomnes:
This should be resolved, however I haven't personally tested.

awohletz:
Confirmed that updating to Electron v6 (currently in beta) fixed the issue for me. E.g. I am able to use node-sqlite3 from an Electron-forked process.

bcomnes:
@awohletz that's great news thank you for the info!

PR #279, opened 2017-03-29
Allow electron process forks of modules that use pre-gyp

bcomnes:
closes https://github.com/mapbox/node-pre-gyp/issues/278

I still need to check to see if this works in built electron apps, but this is fairly close to solving the issue of allowing `childProcess.fork`s in electron of node-pre-gyp modules.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -263,7 +263,8 @@ function get_process_runtime(versions) {
     var runtime = 'node';
     if (versions['node-webkit']) {
         runtime = 'node-webkit';
-    } else if (versions.electron) {
+    } else if (versions.electron || process.env.ELECTRON_RUN_AS_NODE) {
+        // Running in electron or a childProcess.fork of electron
         runtime = 'electron';
     }
     return runtime;
--- a/test/versioning.test.js
+++ b/test/versioning.test.js
@@ -77,6 +77,12 @@ test('should detect runtime for node-webkit and electron', function(t) {
         "node-webkit": '0.37.3',
     };
     t.equal(versioning.get_process_runtime(mock_process_versions3),'node-webkit');
+    var mock_process_versions4 = {
+        "node": '0.8.0',
+    };
+    process.env.ELECTRON_RUN_AS_NODE = 1;
+    t.equal(versioning.get_process_runtime(mock_process_versions4), 'electron');
+    delete process.env.ELECTRON_RUN_AS_NODE;
     t.end();
 });
 


springmeyer:
Thanks for this PR @bcomnes and especially for the detailed description of the problem over at #278. I'm the maintainer of node-pre-gyp and happy to accept this PR once there are tests. I'm not an electron user so I'd love help on how to best set up tests to confirm this is working and ensure it never regresses (which will run on travisci). Are writing unit tests something you could do? Or could you advise on how to test manually and maybe others could help convert those into unit tests?

bcomnes:
👌 I'll put something together today.

bcomnes:
A decent test would be to basically try to load a node-pre-gyp module in a module running in a childProcess.fork of electron.  Would you be opposed to pulling in https://github.com/electron-userland/electron-prebuilt as a dev dependency?

bcomnes:
Yarg, why is that failing

bcomnes:
The other weird thing about this is requiring electron.  Ideally we could pull off the correct electron version from the thing that actually called the fork, rather than an action-at-a-distancey require resolve of a package.json.

bcomnes:
Oh crap, I just pushed some unintentional files up.  Fixing.

bcomnes:
Ok I made a few changes:

- Simplified the electron case of get_process_runtime
- No longer mutate process.versions.  
- The electron version check happens in get_runtime_abi instead
- I opened an issue upstream to electron to expose the version of electron to the forked electron process somehow https://github.com/electron/electron/issues/9058
- Added a mock test to test out the detection of ELECTRON_RUN_AS_NODE

Is that enough testing for this you think? Or do we want to add more of an acceptance / integration test closer to what https://github.com/ballpit/electron-sqlite3-fork-bug/blob/master/main.js tests for?

bcomnes:
@springmeyer any feedback on the tests/current state of the PR?

bcomnes:
Rebased.

bcomnes:
@springmeyer Do you have any feedback on the tests I added?

test/electron/index.js
```diff
@@ -0,0 +1,13 @@
...
+var evaluate = require('../../lib/util/versioning').evaluate;
+var find = require('../../').find;
+var path = require('path')
+
+describe('electron test', function() {
+    it('just pass', function() {
+        assert.equal(1,1);
+        var opts = evaluate(require('../app1/package.json'));
+        console.log(opts)
+        console.log(find(path.resolve(path.join(__dirname,'../app1/package.json'))))
```
springmeyer: This should assert something rather than just `console.log`, correct?

lib/util/versioning.js.orig
```diff
@@ -0,0 +1,317 @@
+"use strict";
```
springmeyer: I presume you added this `lib/util/versioning.js.orig` by accident? Can you remove and rebase?

springmeyer:
@bcomnes - made a request for a few minor changes. If you can address those then I am willing to merge. I admit that I don't fully understand the changes or how they work. So I would love if anyone else that uses electron could weight in.

lib/util/versioning.js.orig
bcomnes: Derp.  Not sure how this snuck in here.  Removing the orig file.

test/electron/index.js
bcomnes: These snuck in as well.  Fixing.

bcomnes: This was a leftover from an attempt to actually perform an integration test with electron, but there are some costs to adding those in like adding additional CI prebuilt downloads.  I decided to just go with the unit tests in `test/versioning.test.js` which seem to cover the issue at play here.  Removing this file for now.

bcomnes:
Ok, cleaned up those two issue, and rebased.  

To TLDR explain the issue: when forking a process in electron, a special ENV var is introduced to the forked process environment (`ELECTRON_RUN_AS_NODE`) and any prebuilt native dependencies need to pick this up in order to find the correct bindings, otherwise the normal checks that node-pre-gyp performs to check if its running electron fail and it returns the vanilla node bindings (or at least this is one approach).  

The only real weirdness here is the electron version resolution.  Ideally this would also come in from the ENV of the forked process, but its not available yet.  I'm currently working with the electron team to get that in (see https://github.com/electron/electron/issues/9058) and will PR that improved method once its available.

bcomnes:
Hrmmm tests are failing now.  Any idea if its because of what I'm doing?

springmeyer:
> Any idea if its because of what I'm doing?

@bcomnes - I think they may be unrelated. Can you rebase against the master branch and see if things are fixed?

bcomnes:
Rebased.  Fingers crossed!

bcomnes:
In a spat of inspiration to make the world better this morning, I squashed and rebased this.

An alternative PR for the same problem is here: https://github.com/mapbox/node-pre-gyp/pull/343

It looks like the difference between the two is ^^ that one requires you to pass in an ENV var yourself, this PR attempts to resolve the correct electron version for you.   Personally, I think this is how it *should* work, but the code is a little more ugly until we can support from upstream electron.

Getting rock solid sqlite support in electron would be amazing, and this would help make that happen.

bcomnes:
All tests passing locally.  CI seems to be failing on unrelated tests 🤔 any ideas?

bcomnes:
@springmeyer any clue whats going on in the failed runs?  Also what happened to travis?

dimalo:
Hey guys, thanks for your work! Please make this merge finally happen!

springmeyer:
Thanks so much @bcomnes. Would love to land this. At the same time I don't have time or ability to review or merge currently. Nor debug travis, but it might be something like https://github.com/mapbox/gzip-hpp/issues/33 which resulted from .org -> .com issues I think.

codecardinal:
@springmeyer I need this PR for sqlite3 too :P

JakeRadMSFT:
@bcomnes it looks like electron made the change you wanted: 
https://github.com/electron/electron/issues/8727

bcomnes:
@JakeRadMSFT The thing that I wanted to remove was https://github.com/mapbox/node-pre-gyp/pull/279/files#diff-314ae8d53157a25e4458ced25e711105R87

https://github.com/electron/electron/pull/16450 adds a `process.versions.electron` which suposedly provides that.  Let me try it out.

bcomnes:
Ok, re-reading this.  Totally untested, but: IF `process.versions.electron` is added into forks, as noted in electron/electron#9058, then this PR is no longer needed.  Since tests arn't passing on master, I dont even think the `process.env.ELECTRON_RUN_AS_NODE` part is needed.

Closing.

awohletz:
Confirmed that updating to Electron v6 (currently in beta) fixed the issue for me. E.g. I am able to use node-sqlite3 from an Electron-forked process.

PR #282, opened 2017-04-03
Clarify config in readme

dszakallas:
I added some clarification about the environment variables and usage of `node_pre_gyp_bucket`


springmeyer:
Thanks. However I'm not understanding `Note that case is taken into account when reading environment variable keys, so be sure to use them verbatim.`. Can you say more about what you mean?

springmeyer:
If this is still a problem, can you re-create a new PR to address the issue? Reason is that a few things have changed in master recently with the v1.0.0 release. So ideally you could test with that and then create a new PR if the docs still need help. Thanks.

Issue #270, opened 2017-03-14
False positive with testbinary command


jfirebaugh:
The `testbinary` command does not seem to validate that the binary actually exists.

1. Create an empty directory and add the following `package.json` file in it:
```
{
  "name": "test",
  "version": "0.0.0",
  "main": "index.js",
  "dependencies": {
    "node-pre-gyp": "^0.6.33"
  },
  "binary": {
    "module_name": "test",
    "module_path": "./lib/",
    "host": "https://mapbox-node-binary.s3.amazonaws.com"
  }
}
```
2. Run `npm install`
3. Run `./node_modules/.bin/node-pre-gyp testbinary`

Expected behavior: get an error saying that the binary doesn't exist.

Actual behavior:

```
node-pre-gyp info it worked if it ends with ok
node-pre-gyp info using node-pre-gyp@0.6.33
node-pre-gyp info using node@4.7.3 | darwin | x64
node-pre-gyp info validate Running test command: '/usr/local/Cellar/node@4/4.7.3/bin/node --eval 'require(\'/Users/john/Development/test-node-pre-gyp/lib/test.node\')''
node-pre-gyp info ok 
```

springmeyer:
Thanks for the report. I can replicate. I've tracked the problem down to this commit: https://github.com/mapbox/node-pre-gyp/commit/7ed0d0860a801a331c136a4d9d73a6622471cb73. Reverting that fixes the problem.

springmeyer:
No response from @edgarsilva - so this is reverted and fixed by #283

PR #283, opened 2017-04-05
Do actually test the binaries

szdavid92:
The second argument has too much quotes so it is passed to node as string, meaning that nothing is actually tested.

### DIFF
--- a/lib/testbinary.js
+++ b/lib/testbinary.js
@@ -62,7 +62,7 @@ function testbinary(gyp, argv, callback) {
         return;
     }
     args.push('--eval');
-    args.push("'require(\\'" + binary_module.replace(/\'/g, '\\\'') +"\\')'");
+    args.push("require('" + binary_module.replace(/'/g, '\'') +"')");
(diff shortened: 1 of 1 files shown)


springmeyer:
@szdavid92 Thanks for working on fixing this! There is a bit more background on the problem at #270. It looks like your changes moves us back to the code that originally worked before https://github.com/mapbox/node-pre-gyp/commit/7ed0d0860a801a331c136a4d9d73a6622471cb73. I was hoping to hear from @edgarsilva on why that change was needed. If we don't hear from @edgarsilva in the next couple of days I'm 👍  on merging this PR.

szdavid92:
maybe it is platform dependent? On Linux and macOS this is the version that is working. Seems a bit silly to me as I would expect the single quotes around the stuff too, very sneaky bug :(. Another option would be to use `exec` instead of `execFile`, which gives more control over quoting.

springmeyer:
Thanks I'll merge this now.

PR #289, opened 2017-05-04
match node-gyp in ca and bad download handling

springmeyer:


springmeyer:
This did not pass and I did not have time to investigate why

nicolasnoble:
What was the issue this was trying to solve?

springmeyer:
I don't remember. I think I was aware of some reason that https://github.com/nodejs/node-gyp/commit/8c4b0ffa503aa385f4186e028542dcedca2efeee might be needed. It might have been https://github.com/mapbox/node-pre-gyp/issues/209, but that turned out to be unrelated.

PR #292, opened 2017-05-16
Search path for npm to use npm's node-gyp

bcomnes:
This fixes issues during weird rebuild scenarios when installing with yarn and rebuilding for electron using electron-builder.

The fix is fairly general though, as it will search the users path for npm, then resolve the node-gyp that it ships.  If npm doesn't exist, it will noop.

I'm not really sure how to test this, other than locally, since its such an environmental codepath.

Related to: https://github.com/yarnpkg/yarn/pull/3240 
and https://github.com/electron-userland/electron-builder/issues/1542

### DIFF
--- a/lib/util/compile.js
+++ b/lib/util/compile.js
@@ -52,6 +52,14 @@ function which_node_gyp() {
     if (existsSync(node_gyp_bin)) {
         return node_gyp_bin;
     }
+
+    try {
+        var maybe_elsewhere = path.join(path.dirname(fs.realpathSync(which('npm'))), 
+                                    '..', 'node_modules/node-gyp/bin/node-gyp.js');
+        if (existsSync(maybe_elsewhere)) {
+            return maybe_elsewhere;
+        }
+    } catch (err) { }
 }
 
 module.exports.run_gyp = function(args,opts,callback) {
--- a/lib/util/compile.js
+++ b/lib/util/compile.js
@@ -4,6 +4,7 @@ module.exports = exports;
 
 var fs = require('fs');
 var path = require('path');
+var which = require('which').sync;
 var win = process.platform == 'win32';
 var existsSync = fs.existsSync || path.existsSync;
 var cp = require('child_process');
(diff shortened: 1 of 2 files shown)


lib/util/compile.js
```diff
@@ -52,6 +52,14 @@ function which_node_gyp() {
     if (existsSync(node_gyp_bin)) {
         return node_gyp_bin;
     }
+
+    try {
+        var maybe_elsewhere = path.join(path.dirname(fs.realpathSync(which('npm'))),
```
springmeyer: Looks like you forgot to require `which`? Also, is it absolutely needed? Would be great to not add more deps.

bcomnes: Oops, given the bizarreness of toolchains and installers and nested dips that are needed to trigger the issue, I had to do work inside of an active node_modules and move the work over once I figured it out.  Adding that requirement now.

An alternative to `which` would be baking in the common path that triggers this problem:
`/usr/local/lib/node_modules/npm` (the default place homebrew moves npm when you install node.). The problem is that users can and do change the basepath that homebrew installs too (don't ask me why 😭) or even introduce other npm installs in other parts of their environment.  Fundamentally though, we are looking for `node-gyp` inside of `npm`'s node_modules folder from the users configured environment.  This is the most direct path to find that as far as I can think of.  [`which`](https://www.npmjs.com/package/which) is a fairly tried and true dependency by issacs that npm itself depends on, so its probably a fairly safe bet.

I also suggested that yarn not silent peer-depend on npm and just ship node-gyp like it used when everything 'just worked' to but I'm waiting to hear back about that idea.

Either way, this adds one other fairly robust system check for node-gyp before giving up.

Daniel15:
I just noticed that `node-pre-gyp` doesn't declare a dependency on `node-gyp` in its `package.json` file. Wouldn't it be safer to make that dependency explicit, rather than implicitly relying on a globally-installed node-gyp?

bcomnes:
I think that would work too. I figured there was a reason for that but I guess I should have asked.

bcomnes:
Im abandoning this PR.

PR #296, opened 2017-05-30
skip testbinary validation for electron

### LINKED ISSUE #295: Electron binaries fail validation in v0.6.35
In a projects with `runtime=electron`, the `node --eval require` from `testbinary.js` results in a library load error because it tries to use node to load the library rather than electron.  This is new in v0.6.35.  At the very least, this test should be skipped if runtime != node until a more robust electron test is available.

springmeyer:
Fixes #295


p120ph37:
Tested - this fixes it for me, thanks.

springmeyer:
Thanks, will merge and release once travis is green.

Issue #300, opened 2017-06-11
Arguments like --dist-url/--nodedir not respected with npm 3.10.10


springmeyer:
## Context

When someone runs `npm install --some-option` to install a module that uses `node-pre-gyp`, then `node-pre-gyp` notices the arguments passed and forwards to `node-gyp` (when a source compile is needed). The same thing is done when `node-pre-gyp` is called directly (rather than via npm). This argument forwarding is critical to be able to control `node-gyp` behavior directly (since modules that use `node-pre-gyp` don't directly call `node-gyp`, by design).

## Problem

This argument passing works in all cases when arguments are passed directly to `node-pre-gyp`. For example `node-pre-gyp build --dist-url=foo` will result in a call to `node-gyp build --dist-url=foo`.

And this argument passing works in all cases when arguments are passed to `npm` and `npm calls` `node-pre-gyp` except on windows with npm 3.10.10. What I've seen is:

 

|             | mac                | linux | windows | 
|-------------|--------------------|-------|---------|
| node v6/npm 3.10.10 | :white_check_mark: | :white_check_mark:  | :x: |
| node v6/npm 2.15.11 (downgraded)       | :white_check_mark: | :white_check_mark:  |  :white_check_mark:  |
| node v4/npm 2.15.11        |:white_check_mark: | :white_check_mark:  |  :white_check_mark:  |
| node v0.10/npm2.15.1       |:white_check_mark: | :white_check_mark:  |  :white_check_mark:  |

springmeyer:
Looks like this bug also impacts npm 5.3.0 / node v8.4.0 on windows: https://ci.appveyor.com/project/Mapbox/node-pre-gyp/build/1.0.625/job/m02cx3p7chk0q4km#L53

And with those versions another problem manifests that looks like:

```
not ok 29 Error: Command failed 'npm install --clang=1 --msvs_version=2015 --fallback-to-build=false'
  ---
    operator: error
    expected: |-
      undefined
    actual: |-
      [Error: Command failed 'npm install --clang=1 --msvs_version=2015   --fallback-to-build=false']
    at: maybeClose (internal/child_process.js:927:16)
    stack: |-
      module.js:491
          throw err;
          ^
      
      Error: Cannot find module 'C:\projects\node-pre-gyp\test\app1\node_modules\node-pre-gyp\bin\node-pre-gyp'
          at Function.Module._resolveFilename (module.js:489:15)
          at Function.Module._load (module.js:439:25)
[...]

Issue #276, opened 2017-03-21
Simplify logic around abi crosswalk


springmeyer:
### Context

When node-pre-gyp started node was at `0.8.x` and `process.versions.modules` had recently been added to node core to allow dynamically knowing the ABI version for a given node version.

node-pre-gyp needs to know this crosswalk in order to allow you to install a module for a given node version different than your host (important for packaging like https://github.com/mapbox/mapbox-studio-classic/blob/89bce13040c0ae3ff7c6c64371a8c428962c8b24/scripts/build-atom.sh#L81-L85). Instead of making you provide the ABI it makes sense to provide your exact node version you are targeting (MAJOR.MINOR.PATCH) and then node-pre-gyp is smart enough to grab the right ABI.

Two constraints existed:

> How to generate the crosswalk between node version and ABI and keep it updated?

In an effort to avoid extra network that might result in a build/install failure for node modules using node-pre-
gyp, I committed to storing the generated crosswalk inside a json file in node-pre-gyp.

> It was unclear at the time how that crosswalk would evolve. Would each node minor version bump the ABI or would each major version change it? Or would major versions get released that did not change the ABI?

Over time what happened is that:

 - During the `0.x` series the minor version bumped the ABI during the even series but during the odd series the ABI got bumped even in patch versions. So, a cross walk of exact major.minor.patch was needed to tame the chaos.
 - During the iojs phase of `1.x` releases the ABI version was bumped during a minor release.
 - After the nodejs series started releasing  `2.x` and above we've seen the ABI only get bumped with new major versions, predictably.

### Problem

This crosswalk is hard to maintain. Every time a node release happens (even a point release) it needs to be updated. And modules depending on node-pre-gyp also need to be updated if they use the `--target` flag and specific a release node version that node-pre-gyp does not know about. Tests in node-pre-gyp assert that it is updated, but this means the node-pre-gyp tests start failing after new releases of node, which is annoying and creates busywork.

### Solution

Here is a potential solution that keeps this functionality working but reduces maintenance burden.

- node-pre-gyp is modified to require an extra flag that points to an json file on the on the filesystem.
- That file would be in the format of https://nodejs.org/download/release/index.json
- it would be expected that users needing the `--target` flag would download https://nodejs.org/download/release/index.json and supply the path to it to node-pre-gyp
- On travis this would be tested such that every travis build would download https://nodejs.org/download/release/index.json and ensure it works to use.

bruce-one:
Hmm...

[...]

PR #301, opened 2017-06-11
Refactor tests / port to tape

springmeyer:
While debugging #300 I suspected that the mocha tests might be behaving unpredictably. I wondered if porting to tape would resolve the issues in #300. They did not, but moving to `tape` seems ideal anyway (given https://macwright.org/2014/03/11/tape-is-cool.html). so I forged ahead on these.

Also solved in this PR along the way:

 - Solves one part of #276:
    - Because `After the nodejs series started releasing 2.x and above we've seen the ABI only get bumped with new major versions, predictably.` we now assume the ABI even when the exact minor.patch version is not stored in the abi crosswalk. This will avoid it needing to be updated as frequently (only for new major versions).
 - Applied better workaround in the appveyor.yml for https://github.com/mapbox/node-pre-gyp/issues/209
 - Started returning a non-zero error code with invalid usage (and added test for this).
 - Solves #302 and #300 by downgrading npm to avoid npm bugs

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,16 +17,15 @@ install:
   - ps: Install-Product node $env:nodejs_version $env:Platform
   - ps: Set-ExecutionPolicy Unrestricted -Scope CurrentUser -Force
   - npm config get
-  # make node >= 4.x x86 builds work
-  # https://github.com/mapbox/node-pre-gyp/issues/209#issuecomment-217690537
-  #- IF /I "%PLATFORM%" == "x86" IF "%nodejs_version:~0,1%"=="4" npm config set -g cafile=package.json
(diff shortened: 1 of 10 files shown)


isaacs:
Unless you're planning on running tests in the browser (which node-pre-gyp _probably_ isn't going to do), node-tap might be a better fit.  Then you get the runner, reporter, coverage, and test framework all in one.

ajmas:
Any status update?

springmeyer:
Going to stick with tape for the moment. I'm short on time and it was not immediately obvious to me how to port https://github.com/mapbox/node-pre-gyp/blob/07704da675bfa0bdbae0c9f3bcdd9575099a8377/test/build.test.js#L37-L44 over.

PR #311, opened 2017-08-08
Chakracore support

jackTheRipper:
**Do**

Add support for the Node-Chakracore.

If the version of NodeJS used with `node-pre-gyp` is not using v8, it creates an archive where we add `-chakracore` to the `abi` name.

Fix unit-tests as well for Travis-CI.


jackTheRipper:
I am missing the AppVeyor configuration for running the tests for Node-8 Chakracore. I will add it.

test/build.test.js
```diff
@@ -19,6 +19,8 @@ if (process.platform === 'win32') {
 process.env.PATH = cmd_path + sep + process.env.PATH;
 process.env.NODE_PATH = path.join(__dirname,'../lib/');
 
+var jsEngine = process.jsEngine ? '-' + process.jsEngine : '';
```
kunalspathak: >process.jsEngine ? '-' + process.jsEngine : ''; [](start = 15, length = 47)

I would just use `process.jsEngine || 'v8'` unless you have reason of doing it this way.

test/build.test.js
```diff
@@ -98,7 +102,7 @@ function on_error(err,stdout,stderr) {
     throw new Error(msg);
 }
 
-var current_version = process.version.replace('v','');
+var current_version = process.version.replace('v','').replace(jsEngine, '');
```
kunalspathak: >.replace(jsEngine, ''); [](start = 53, length = 23)

why is this needed?

test/versioning.test.js
```diff
@@ -49,15 +50,23 @@ describe('versioning', function() {
 
     it('should detect abi for custom node target', function() {
         var mock_process_versions = {
-            "node": '0.10.0',
-            "modules": '11',
+            "node": (jsEngine) ? '8.1.2' : '0.10.0',
```
kunalspathak: >jsEngine [](start = 21, length = 8)

People can build node-v8 from node-chakracore source in which case `process.jsEngine` will be `v8` . If we want strict chakracore check then do `jsEngine === 'chakracore'` instead.

scripts/latest-by-abi.js
```diff
@@ -4,13 +4,14 @@ var semver = require('semver');
 var data = require('../lib/util/abi_crosswalk.json');
 
 var targets = {};
+var jsEngine = process.jsEngine ? '-chakracore' : '';
```
kunalspathak: >jsEngine [](start = 23, length = 8)

same here. change to `process.jsEngine === 'chakracore' ?? '-chakracore' : ''`

package.json
```diff
@@ -45,7 +45,7 @@
     "mocha": true
   },
   "scripts": {
-    "pretest": "retire -n && jshint test/build.test.js test/s3_setup.test.js test/versioning.test.js",
+    "pretest": "retire -np && jshint test/build.test.js test/s3_setup.test.js test/versioning.test.js",
```
kunalspathak: >p [](start = 25, length = 1)

intended change?

lib/util/versioning.js
```diff
@@ -17,6 +17,8 @@ if (process.env.NODE_PRE_GYP_ABI_CROSSWALK) {
     abi_crosswalk = require('./abi_crosswalk.json');
 }
 
+var jsEngine = process.jsEngine ? '-' + process.jsEngine : '';
```
kunalspathak: >jsEngine [](start = 4, length = 8)

as mentioned in my previous comments, i would do explicit check for `chakracore` here.

kunalspathak:
appveyor doesn't have native support for `node-chakracore` installation. I used `nvs` to achieve this. See https://github.com/expressjs/express/pull/3251

test/build.test.js
jackTheRipper: It was in order to not to create a breaking change. If we add the suffix `v8` everywhere it could break the versioning of the modules (such as `fsevents`)

test/versioning.test.js
jackTheRipper: Ok, I get your point. I will fix it

package.json
jackTheRipper: Yes, because of the security issue with Growl. the `-p` checks only the dependencies

appveyor.yml
```diff
@@ -15,8 +17,12 @@ shallow_clone: true
 install:
   - ps: Install-Product node $env:nodejs_version $env:Platform
```
kunalspathak: I think you need to remove these lines because `nvs` will take care of this.

jackTheRipper: oh ok. Doing it now

test/build.test.js
jackTheRipper: Because otherwise it tries to get `version-${jsEngine}` but it fails with NVM

test/versioning.test.js
jackTheRipper: Done line#6

jackTheRipper:
Regarding the appveyor build I can't do much for fixing the error

jackTheRipper:
Regarding the appveyor build I can't do much for fixing the error

appveyor.yml
```diff
@@ -1,37 +1,38 @@
...
   - npm install node-gyp@3.x
-  - node --version
-  - npm --version
   - node -e "console.log(process.arch);"
   - IF /I "%PLATFORM%" == "x64" set PATH=C:\Python27-x64;%PATH%
   - IF /I "%PLATFORM%" == "x86" SET PATH=C:\python27;%PATH%
   - IF /I "%PLATFORM%" == "x64" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
   - IF /I "%PLATFORM%" == "x86" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
   - npm install
-  - npm test
```
kunalspathak: just to verify the cause, could you undo this change? Bring back `npm test` under `install` step?

PR #313, opened 2017-09-08
Port to node-tap

springmeyer:
PR to experiment with node-tap against master.

refs https://github.com/mapbox/node-pre-gyp/pull/299#issuecomment-327963987

PR #314, opened 2017-09-08
Use tap

isaacs:
Land after #299
Supersedes #313


springmeyer:
we've been doing okay with tape, so not planning on working to update/land this.

PR #319, opened 2017-09-22
Keep compatibility with node v0.10.x

springmeyer:
My plan is to keep compatibility with node v0.10.x in the 0.x series and bump to v1.x (dropping node v0.10.x support) after the merge of #299

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,10 @@ addons:
 
 env:
   matrix:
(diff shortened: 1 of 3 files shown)


springmeyer:
> This is needed due to request/request#2772

@aikar see above.

However now v0.7.0 of node-pre-gyp has dropped node v0.10.x support, so now we're back on more recent request.

PR #320, opened 2017-09-26
Update request to 2.82.0

godspeedelbow:
So, that transitive dep `ajv` gets bumped from `4.x` to `5.x`

see: https://github.com/request/request/issues/2652


godspeedelbow:
```
/Users/travis/build/mapbox/node-pre-gyp/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
gyp ERR! UNCAUGHT EXCEPTION 
gyp ERR! stack SyntaxError: Use of const in strict mode.
```

Seems like a transitive dependency of request, `boom`, is not sticking to ES5.

JLHwung:
Suggest update request to 2.83.0 due to https://github.com/request/request/pull/2776.

JLHwung:
As long as test passes on node >= 4 and node = 0.10 was added as an attempt to fix a bug, I think it is safe to merge this PR.

godspeedelbow:
it fails only on 0.10

igorlima:
Is it possible to have an update on this 🙏

PR #321, opened 2017-09-28
Update dependency on request to address ReDoS

dancrumb:
Latest version of `request` pulls in `tough-cookie` which recently address a ReDoS vulnerability: salesforce/tough-cookie#92


dancrumb:
The CI failures appear to be unrelated to this PR and related to a Node 0.10 issue

HowWeRollingham:
This could be resolved by enabling the use of `const` and `let` in node 0.10.48 with the --harmony flag

acoard:
I assumed this would be a simple version bump from 2.81.0 to 2.83.0, but if I understand right after you bumped the version the tests failed.

Any update on this? And if you could provide a general timeline (hours/days/weeks/months/never) I'd really appreciate it.  We're required to get these vulnerabilities out of our code before we ship. So if this will be a while (weeks+) I'd appreciate knowing so we can plan accordingly.  Hope I'm not asking too much, and I just wanted to say thank you for writing software our project depends on. 👍

dancrumb:
This looks like an issue with `request` having a dependency on code that uses `const` (as indicated by @HowWeRollingham)

Since node 0.10 is EOL (https://github.com/nodejs/Release), I'd recommend just dropping support for it...
Otherwise, you'll need to co-ordinate with the `request` folks, but I'd be surprised if they want to spend time supporting an obsolete level of node.

springmeyer:
Addressed by #347

PR #324, opened 2017-10-20
Don't fix versions in dependencies

gfx:
To optimize `node_modules/`, it's better not to fix dependency versions.


springmeyer:
Done in #347

Issue #325, opened 2017-10-21
Variable for libc type (e.g. differentiate Alpine and Debian)


core-process:
We are building binaries for Alpine and Debian, therefore we need to differentiate libc variants (glibc vs. musl, not versions!). Therefore we need to differentiate the binaries we are producing.

Is there already a way to accomplish this?

core-process:
I understand there are many scenarios which require yet another differentiator. Also I understand there is the toolset-parameter. Still I think the toolset parameter is not a solution, because it requires external intervention during an "npm install" of the library (if I am not mistaken).

I think it would be a good idea to allow a library specific toolset-detection script, which gets executed and will be considered during the compilation of the version-variable. This would enable library developers to make sure the right binary gets installed without user intervention.

core-process:
@springmeyer How do you feel about this?

springmeyer:
@core-process my feelings are:

 - This issue is important: I want to support/offer/document a good, maintainable solution for better supporting binaries on alpine linux/(linux systems with an alternative libc).
 - It feels like the build could be easily automated by adding a .travis/circleci job that compiles inside docker with an alpine linux image
 - The `--tooset` arg, being arbitrary, could be used to accomplish the correct versioning of the binaries now: so there is nothing inside node-pre-gyp that should prevent this from working. But as you say, it would not be fully automatic: You would need to manually provide `--tooset=musl` (where `musl` is arbitrary keyword) when building and also manually provide `--tooset=musl` when installing.
 - Writing a script to automatically detect the libc version might be easy in a one-off way, but very difficult to write in a way that would be robust enough to include in node-pre-gyp.


So, it feels like the next step would be to create a new module, in JS, that is designed to detect the libc version. It would need to be tested on a lot of different linux systems. It would need to be robust. Is this something you'd be interested in working on?

Once done, that module could be potentially used in node-pre-gyp to make the versioning  automatic.

If that milestone was achieved, then we could consider adding a new versioning `{token}` called `{libc}`. Users could opt-in to supporting this by putting `{libc}` in their versioning string by manually providing a custom `package_name` or `module_path` property in their `package.json`. So, by default binaries would not be versioned on libc (to keep back compatibility). But a dev wanting to opt-in could do something like:

```json
    "binary": {
        ...
        "module_path": "./lib/binding/{configuration}/{libc}/{name}",
        ...
    },
```

core-process:
@springmeyer Thanks for your response. 

[...]

PR #326, opened 2017-10-29
add libc detection and {libc} versioning token

core-process:
Improve support for alternate libc libraries, especially `musl` (Alpine), by adding a `{libc}` versioning token. Related to https://github.com/mapbox/node-pre-gyp/issues/325.

Note: A unit test is included (see last test in `test/versioning.test.js`).


core-process:
See https://github.com/lovell/detect-libc/pull/9

core-process:
@springmeyer requesting review and merge :-)...

springmeyer:
This is incredible work and looks perfect. I will merge now. Will plan to do a release asap.

springmeyer:
released in node-pre-gyp@0.6.39

core-process:
Awesome, thanks! :-)

core-process:
@lovell just fyi: your library is integrated now... :-)

rmg:
This has broken my dockerized app which has none of the binaries in the production image that `detect-libc` uses to infer the libc version :-(

springmeyer:
@rmg thanks for the report - my assumption is that this would be harmless for everyone that did not opt in. Maybe I missed something - what exact error are you seeing?

core-process:
@springmeyer related to https://github.com/lovell/detect-libc/issues/10

rmg:
@core-process thanks for linking it back, I got side-tracked by hunting down the cause. Good news is I found it and there's a pretty trivial workaround I can use on my end.

core-process:
@rmg thanks for your input. Still we want to handle this specific case in detect-libc too. I am glad you found a workaround meanwhile!

springmeyer:
Thanks for the cross link @core-process - I'm still confused however: does anything break if a users is not using the `{libc}` variable? Or is the only issue that trying to get the `{libc}` fails due to lovell/detect-libc#10?

core-process:
@springmeyer `{libc}` resolves to `"unknown"` in this specific case.

rmg:
The actual module I was using was `grpc`, which _does_ use the `{libc}` in its binary name template.

core-process:
@rmg Yes of course it breaks modules which are using this token. @lovell just published the new version of `detect-libc` at npmjs.com and according to our integration test suite your case should work too now. Please let as know if we resolved the issue.

rmg:
I was mostly pointing it out for the benefit of Google :-)

PR #342, opened 2018-01-08
update `request` from 2.81.0 to 2.83.0

jinwoo:
The main purpose of this PR is for removing the (transitive) dependency
on `jsonify`, which has a problematic license for us (Public Domain).

`request` upgraded `har-validator` from ~4.2.1 to ~5.0.3 and it
eventually removed the dependency on `jsonify`.

Hope you don't mind this change.


jinwoo:
Tests fail with Node 0.10. And I also notice that there was a previous attempt for this: https://github.com/mapbox/node-pre-gyp/pull/320.

Is it really required to support Node 0.10? It doesn't seem ideal to fix the version of `request` to 2.81.0.

springmeyer:
Done in #347

PR #343, opened 2018-01-12
get correct electron version when addon is running by child_process fork

Tomyail:
Because i can not direct get electron version in child_process, see [this](https://github.com/electron/electron/issues/9058). so `ELECTRON_RUN_AS_NODE ` is not [work](https://github.com/mapbox/node-pre-gyp/pull/279).
We can pass some process env to fork method like this:
```
// run in electron main process
const env = {...process.env,ELECTRON_VERSION:process.versions.electron};
const p = require( 'child_process').fork(`${ __dirname }/child_process.js`,[],{env:env});
```

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -79,7 +79,7 @@ function get_runtime_abi(runtime, target_version) {
     if (runtime === 'node-webkit') {
         return get_node_webkit_abi(runtime, target_version || process.versions['node-webkit']);
     } else if (runtime === 'electron') {
-        return get_electron_abi(runtime, target_version || process.versions.electron);
+        return get_electron_abi(runtime, target_version || process.versions.electron || process.env.ELECTRON_VERSION);
     } else {
         if (runtime != 'node') {
             throw new Error("Unknown Runtime: '" + runtime + "'");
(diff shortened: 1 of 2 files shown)


bcomnes:
I made a similar PR, but we were never able to get unrelated tests from failing in https://github.com/mapbox/node-pre-gyp/pull/279

Allowing It–to-just-work in all the ways to spawn in electron would be fantastic, but I'm no longer working on that problem.  Good luck!

dsagal:
This PR looks useful to me too! Any chance it could get in?

BenjaminVerble:
I need this fix as well

d4tocchini:
I need this as well.

any reasons why it's not yet merged?

JakeRadMSFT:
@Tomyail I think you can get the version now: https://github.com/electron/electron/pull/16450

Tomyail:
@JakeRadMSFT Thank you for remind， I will close the pr.

devinsm:
@Tomyail @JakeRadMSFT I got here from [mapbox/node-pre-gyp#278](mapbox/node-pre-gyp#278), since this pull request seemed to be solving the same issue and was referenced there. I'm trying to use node-sqlite3 from a forked child_process in electron, but ran into the same issues described in [mapbox/node-pre-gyp#278](mapbox/node-pre-gyp#278). How does having access to the electron version in the child_process fix that issue?

PR #350, opened 2018-03-10
Replace request with needle

### LINKED ISSUE #340: Thoughts on replacing 'request' library
It seems the only place the `request` library is used is in `lib/install.js`, where it's only used for a simple task.
The `request` library is 52 packages, which take up around 4.22 MB, which seems unjustified for the small one-time usage that it gets included for. Would a PR be welcome to replace it with a lightweight alternative, such as [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch) or [simple-get](https://www.npmjs.com/package/simple-get) or something similar?

cktang88:
Closes #340

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -58,7 +58,7 @@ function download(uri,opts,callback) {
       }
     }
     try {
-        req = require('request')(requestOpts);
+        req = require('needle').get(requestOpts.uri, requestOpts);
     } catch (e) {
         return callback(e);
     }
--- a/package.json
+++ b/package.json
@@ -22,11 +22,11 @@
   "dependencies": {
     "detect-libc": "^1.0.2",
     "mkdirp": "^0.5.1",
+    "needle": "^2.2.0",
     "nopt": "^4.0.1",
     "npm-packlist": "^1.1.6",
     "npmlog": "^4.0.2",
     "rc": "^1.1.7",
-    "request": "2.83.0",
     "rimraf": "^2.6.1",
     "semver": "^5.3.0",
     "tar": "^4"


springmeyer:
Looks good, thanks for putting this together. Only test that is failing is unrelated. For the sake of our future selves, can you share why you chose `needle`? After that I'll plan to merge.

cktang88:
@springmeyer I choose `needle` for the following reasons (in no particular order):
1. Had similar configuration option names with `request`, so that few changes to `/lib/install.js` would be needed (specifically, needed to support proxy and HTTP headers)
2. Seems to be widely used, reliable, and actively developed (1.28 million downloads/month)
3. Very lightweight, few dependencies

ninox92:
What about `got`

LinusU:
Name | Downloads <sup>1</sup> | Bundle size <sup>2</sup>
----- | ----: | ----:
`simple-get` | 1,747,443 | **80kB**
`got` | 11,009,114 | 156kB
`needle` | 1,231,722 | 556kB
`request` | **39,390,626** | ??? <sup>3</sup>

1: from [npmjs.com](https://www.npmjs.com/)
2: from [bundlephobia.com](https://bundlephobia.com/)
3: it timed out, so probably quite large 😄

cktang88:
If you consider actual disk size, on my Win10 (NTFS file system), `needle` is 635kb, while `got` is 564kb, so a difference of only 71kb. Although I admit `simple-get` is a lot smaller. Also, for what it's worth, `needle` only installs 5 packages, while `got` installs 44.

cktang88:
Hmmm.... if we are going to have a discussion about all the possible replacement options, I believe [axios](https://www.npmjs.com/package/axios) should also get a mention. It's only 13.6kb according to [bundlephobia.com](https://bundlephobia.com/result?p=axios@0.18.0) which beats all three options mentioned thus far, and has 5.8mil downloads/month.

LinusU:
The problem with adding `axios` and `superagent` to my list is that they have an alternative browser implementation... Listing sizes from bundlephobia.com is actually not a good idea at all, it was just the first tool I found 😄 

The best would be to try and install them in a clean directory and then measuring disk size, hopefully I'll get some time for that this evening

LinusU:
Okay, here is a list using actual install sizes, which is a much better measurement for our use case:

Name | Downloads <sup>1</sup> | Installed size <sup>2</sup>
----- | ----: | ----:
`axios` | 5,802,061 | 588kB
`got` | 11,009,114 | 1260kB
`needle` | 1,231,722 | 888kB
`request` | **39,390,626** | 5800kB
`simple-get` | 1,747,443 | **140kB**
`superagent` | 6,061,245 | 1852kB

1: from [npmjs.com](https://www.npmjs.com/)
2: from running `echo {} > package.json; npm i $NAME; du -sk .` in an empty directory

ninox92:
@linusU are these production installs?

LinusU:
@ninox92 yes 👍

springmeyer:
Thanks for the extra details and discussion here. I like the idea of going out with needle now since it ranks well for size but does not require code changes.

springmeyer:
v0.9.0 is now published with this PR included.

Issue #355, opened 2018-03-18
Crash build


marcobraghim:
`$ node-pre-gyp build --runtime=node-webkit --target=0.13.0 --target_arch=x64`
It produces:

```
/usr/lib/node_modules/node-pre-gyp/lib/util/napi.js:103
	if (package_json.binary.napi_versions) { // remove duplicates
	                        ^

TypeError: Cannot read property 'napi_versions' of undefined
    at Object.module.exports.get_napi_build_versions (/usr/lib/node_modules/node-pre-gyp/lib/util/napi.js:103:26)
    at Object.module.exports.expand_commands (/usr/lib/node_modules/node-pre-gyp/lib/util/napi.js:82:43)
    at Run.parseOpts [as parseArgv] (/usr/lib/node_modules/node-pre-gyp/lib/node-pre-gyp.js:138:20)
    at Object.<anonymous> (/usr/lib/node_modules/node-pre-gyp/bin/node-pre-gyp:24:6)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
```

jschlight:
Does your `package.json` file have a `binary` property which is required?

jschlight:
I've verified on my development system that the reported error occurs when the `package.json` file is missing the `binary` property. I have code ready for a pull request that addresses this.

marcobraghim:
Well, thank you, but I still don't know what it exactly means =/
Sorry for take too much time to reply abou that...

jschlight:
Hello @marcobraghim. It looks like your `package.json` file does not include the `binary` property which is required by `node-pre-gyp`. I have updated `node-pre-gyp` to display a more informative message when the `binary` property is missing, as it likely is in your case.
There is more information about the `binary` property here: https://github.com/mapbox/node-pre-gyp#configuring

marcobraghim:
Got this, thank you.

springmeyer:
solved by #357

PR #357, opened 2018-03-22
Better handle a package.json file without a binary property

inspiredware:
Fixes issue #355

The code now issues a more informative error message when the `package.json` file does not have `binary` property.

Also, a couple of documentation tweaks.


springmeyer:
Thanks @inspiredware 🙏

Issue #359, opened 2018-03-28
Regression: binary installation does not follow redirects


allenluce:
I've noticed this and had a user tell me about the same thing. When doing an `npm install mmap-object`:

```
> mmap-object@1.3.7 install /private/tmp/x/node_modules/mmap-object
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! Tried to download(302): https://github.com/allenluce/mmap-object/releases/download/1.3.7/mmap-object-v1.3.7-node-v48-darwin-x64.tar.gz
node-pre-gyp ERR! Pre-built binaries not found for mmap-object@1.3.7 and node@6.3.0 (node-v48 ABI, unknown) (falling back to source compile with node-gyp)
```
This is a new problem, I've only seen it within the last week or two.

allenluce:
Looks like switching to needle (as part of #350) caused this. Needle doesn't follow redirects by default (as request does).

allenluce:
This is pretty important as my module as it requires C++11 and Boost to be installed for compilation. That's [not always so straightforward for everyone](https://github.com/allenluce/mmap-object/issues/14).

royaltm:
Yes, and currently using `node-pre-gyp-github` is impossible now, as github is redirecting it's release archives now to S3, e.g.:

`curl -D - https://github.com/royaltm/node-murmurhash-native/releases/download/v3.2.3/mu
rmurhash-v3.2.3-node-v46-linux-x64.tar.gz

HTTP/1.1 302 Found
Server: GitHub.com
Date: Thu, 29 Mar 2018 16:10:50 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Status: 302 Found
(...)
Location: https://github-production-release-asset-2e65be.s3.amazonaws.com/(...)```

springmeyer:
Fixed per https://github.com/mapbox/node-pre-gyp/pull/361#issuecomment-380947265. Thank you @allenluce

allenluce:
Thanks for the merge!

fireairforce:
So how to solve this question?

PR #361, opened 2018-03-28
Have needle follow redirects

allenluce:
Redirect following for binary installations broke with the [switch to needle](https://github.com/mapbox/node-pre-gyp/issues/350). This replaces the functionality.

Not having a redirect means (at least in my case) that pre-built binaries that are registered as part of a Github release cannot be downloaded.

This addresses #359

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -32,7 +32,8 @@ function download(uri,opts,callback) {
         uri: uri.replace('+','%2B'),
         headers: {
           'User-Agent': 'node-pre-gyp (v' + npgVersion + ', ' + envVersionInfo + ')'
-        }
+        },
+        follow_max: 10,
     };
 
     if (opts.cafile) {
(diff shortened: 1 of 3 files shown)


lib/install.js
```diff
@@ -105,7 +106,10 @@ function place_binary(from,to,opts,callback) {
         });
 
         req.on('response', function(res) {
-            if (res.statusCode !== 200) {
+           if (res.statusCode === 302) {
```
pimterry: You probably want to allow other 30x codes here too I think.

springmeyer:
Thanks for catching this regression. Before merging this we'll need some kind of test to make sure it is actually working + does not regress again. Could you add a test?

lib/install.js
allenluce: I've changed it to check for a Location: header instead.

allenluce:
I can add a test, how would you like such a thing structured? Typically I'll use nock or the like to simulate some real http action. Doesn't look like there's anything like that in the current suite....

allenluce:
FYI, I've just now noticed the error and have been able to repro it (using Node 4 under OSX, Node 6 on my box seems to work fine). I'll be able to tackle this in a couple hours.

royaltm:
I affects all node versions, all systems.

allenluce:
Is there something keeping this from being merged?

springmeyer:
Thanks for this work, it is now included in `node-pre-gyp@0.9.1`

PR #366, opened 2018-04-19
Log warnings instead errors when falling back

### LINKED ISSUE #344: Is "Pre-built binaries not found for ...." really an error?
I guess this is more of a discussion point rather than a bug per-se, but please hear me out :-)

node-pre-gyp throws an error if it's requested to download a binary that doesn't exist for a specific platform. It then throws another error noting that it's falling back to compiling sources.

eg:
```
node-pre-gyp ERR! Tried to download(403): https://mapbox-node-binary.s3.amazonaws.com/sqlite3/v3.1.13/node-v48-linux-arm.tar.gz 
node-pre-gyp ERR! Pre-built binaries not found for sqlite3@3.1.13 and node@6.12.3 (node-v48 ABI) (falling back to source compile with node-gyp) 
```
What the software is doing is expected, but it's not actually an "error" unless the compile fails - so misleading.

In my opinion the errors above should be logged as warnings, meaning that users not-so-familiar with Node.js don't get confused by messages in logs which can be safely ignored

Would appreciate thoughts on this!

Cheers
Dave

xzyfer:
Fixes #344

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -142,8 +142,8 @@ function print_fallback_error(err,opts,package_json) {
         full_message = "Pre-built binaries not found for " + package_json.name + "@" + package_json.version;
         full_message += " and " + opts.runtime + "@" + (opts.target || process.versions.node) + " (" + opts.node_abi + " ABI, " + opts.libc + ")";
         full_message += fallback_message;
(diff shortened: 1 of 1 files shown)


springmeyer:
Thank you @xzyfer, merging now and this will be part of the upcoming `v0.8.1` release.

PR #371, opened 2018-04-26
Allow for single-level module_path when packing

### LINKED ISSUE #364: Pre-built binaries not installable for ... and node@8.11.1 (node-v57 ABI, unknown) (falling back to source compile with node-gyp)
Hi guys,

  since yesterday I cannot retrieve my pre-built binaries (posted on amazon S3) using "latest" node-pre-gyp.

I suspect a regression in node-pre-gyp..

Here is what I get:

```
Pre-built binaries not installable for xxxxx and node@8.11.1 (node-v57 ABI, unknown) (falling back to source compile with node-gyp)
```

Thanks to investigate asap, this is a blocking issue.

allenluce:
Prior to 0.7.1, a module_path of `./lib` would result in a package containing files like this (using `test/app8` as an example):

    lib/app8.node

The switch to tar 3 in 0.7.1 causes the same package to end up like this:

    package.json
    app8.cc
    binding.gyp
    index.js
    README.md
    lib/app8.node

which causes a problem during remote binary installation:

    node-pre-gyp ERR! Pre-built binaries not installable for PACKAGE@0.0.1 and node@8.3.0 (node-v57 ABI, glibc) (falling back to source compile with node-gyp)
    node-pre-gyp ERR! Hit error ENOTDIR: Cannot cd into '..dirs/package/lib'

This commit fixes this problem and allows for single-level module_path values.

Fixes #364

### DIFF
--- a/lib/package.js
+++ b/lib/package.js
@@ -33,15 +33,18 @@ function _package(gyp, argv, callback) {
             return true;
         };
         mkdirp(path.dirname(tarball),function(err) {
-            from = path.dirname(from);
             if (err) return callback(err);
             packlist({ path: from }).then(function(files) {
+                var base = path.basename(from);
+                files = files.map(function(file) {
+                    return path.join(base, file);
+                });
                 tar.create({
                     portable: true,
                     gzip: true,
                     onentry: filter_func,
                     file: tarball,
-                    cwd: from
+                    cwd: path.dirname(from)
                 }, files, function(err) {
                     if (err)  console.error('['+package_json.name+'] ' + err.message);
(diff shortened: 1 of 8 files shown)


springmeyer:
@allenluce thank you for this. Looks like there are some minor failures (https://travis-ci.org/mapbox/node-pre-gyp/jobs/371702547#L805) when this test is run on different operating systems. How about, instead of checking actual file names, we just check the # of files added?

allenluce:
I always seem to forget this bit! Let me know if the last commit will do it for you.

PR #372, opened 2018-04-27
updates to travis matrix

mapsam:
This updates the travis matrix to:

* remove the RETIRE warnings from CI builds (but keep node v6 tests)
* remove Node v4 testing (Node v4 is EOL on April 30th https://github.com/nodejs/Release)
* add Node v10 builds

cc @springmeyer for review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,7 @@ addons:
 
 env:
   matrix:
-   - NODE_VERSION="5"
    - NODE_VERSION="6"
-   - NODE_VERSION="7"
    - NODE_VERSION="8"
    - NODE_VERSION="9"
    - NODE_VERSION="10"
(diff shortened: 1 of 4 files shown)


springmeyer:
@mapsam #375 (which I just merged in here) fixes the windows issues such that npm no longer needs to be downgraded. The race conditions remain it appears.

mapsam:
@springmeyer excellent! That's really settling. Going to work on getting tests 🍏 for travis now

springmeyer:
> Going to work on getting tests 🍏 for travis now

👍 one shortcut I'm very comfortable with: merging now and ensuring things pass in master. Reasoning is that I think the race condition will be lessened (or gone?) when we are not running two build jobs (like are run for PRs). Still need to solve the race condition, but that can be after release since it does not impact node-pre-gyp users (and is only a test issue).

Issue #367, opened 2018-04-23
Error: Completion callback never invoked


mingyuelou1997:
E:\TEST\sqlite3>npm install sqlite3

> sqlite3@4.0.0 install E:\TEST\sqlite3\node_modules\sqlite3
> node-pre-gyp install --fallback-to-build

node-pre-gyp ERR! Completion callback never invoked!
node-pre-gyp ERR! System Windows_NT 6.1.7601
node-pre-gyp ERR! command "D:\\nodejs\\node.exe" "E:\\TEST\\sqlite3\\node_module
s\\sqlite3\\node_modules\\node-pre-gyp\\bin\\node-pre-gyp" "install" "--fallback
-to-build"
node-pre-gyp ERR! cwd E:\TEST\sqlite3\node_modules\sqlite3
node-pre-gyp ERR! node -v v8.9.3
node-pre-gyp ERR! node-pre-gyp -v v0.9.0
node-pre-gyp ERR! This is a bug in `node-pre-gyp`.
node-pre-gyp ERR! Try to update node-pre-gyp and file an issue if it does not he
lp:
node-pre-gyp ERR!     <https://github.com/mapbox/node-pre-gyp/issues>
npm ERR! code ENOSYS
npm ERR! errno ENOSYS
npm ERR! syscall kill
npm ERR! kill ENOSYS

netbuffer:
face the same issue

springmeyer:
> Completion callback never invoked

@mapsam also saw this recently on travis. I'm going to rename the title of this ticket `Completion callback never invoked`.

mapsam:
Indeed - here's an OSX build with node-pre-gyp tests failing (irregularly) with the same error: https://travis-ci.org/mapbox/node-pre-gyp/jobs/372259045

skyh24:
how to solve this？？？？

TaikiAkita:
Really sucks when I'm building node-webrtc...
Please solve this FV*King problem immediately!!!!

IgorMaineti:
Could someone look into this issue please? I'm facing the same problem

Rosso84:
[...]

PR #377, opened 2018-05-07
change needle "error" event to "err" event, fix "Completion callback never invoked!"

medns:
fixbug: if network unreachable, will crash with "Completion callback never invoked!"
needle "error" event called "err".

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -94,7 +94,7 @@ function place_binary(from,to,opts,callback) {
             extractCount++;
         }
 
-        req.on('error', function(err) {
+        req.on('err', function(err) {
             badDownload = true;
             return callback(err);
         });


mapsam:
Oh wow, good eye @medns! Confirmed `err` is correct in the needle README https://github.com/tomas/needle#event-err - this must be leftover from when we were using request.js

I'm going to test your fix out in conjunction with resolving #367 to confirm it helps. Will report back.

mapsam:
Looks like this is helping out, and we're seeing `Error: socket hang up` come out of needle instead of a vague "Completion callback never invoked!" error. Going to dig into why the socket is timing out (currently theory is just concurrent travis jobs or some sort of limited http sockets on travis/remote boxes), and if we can do anything about increasing the timeout with needle.

mapsam:
https://travis-ci.org/mapbox/node-pre-gyp/jobs/392625721 shows a single failure:

```
[NEEDLE DEBUG] socket timeout??: 'open'
```

And this refers to the socket timing out before it receives a response. According to the needle docs we can change this timeout parameter with the `open_timeout` option:

> Returns error if connection takes longer than X milisecs to establish. Defaults to 10000 (10 secs). 0 means no timeout.

I'll make these changes in a separate PR, but this one is 👍ready to rock. Thanks so much @medns, this will remove a lot of headaches for folks in #367 😄

Issue #383, opened 2018-05-23
libc variable should default to unknown for win32 target_platform?


mark-buer:
## TLDR;
I claim that the `libc` variable makes no sense for windows targets and should default to "unknown" when installing prebuilts on windows _and_ when installing prebuilts for windows on linux.

## Long Story
When invoking node-pre-gyp from win32, `libc` will default to the value "unknown".
When invoking node-pre-gyp from a linux platform (i.e. docker), but targeting win32, `libc` will default to whatever value is detected for the host. Thus, if a native package includes `libc` within the `package_name` template, node-pre-gyp will attempt to download the wrong prebuilt binary.

A real world example of "installing prebuilts _for_ windows _on_ linux": electron-builder allows one to build a windows app distribution from a docker instance if all native modules have win32 prebuilts available. This mostly works, except for the case where the native module uses the `libc` variable within the `package_name` template. An example of such a package is `grpc-node`, which is used by many packages such as `firebase` for example.

Thus, should node-pre-gyp default the `libc` variable to unknown when targeting win32?

Or, is it the responsibility of each and every client of node-pre-gyp to ensure that the `target_libc` option is set to "unknown" for the edge case of cross-installing dependencies?

mark-buer:
Further, perhaps "unknown" should be the default value of libc for all non-linux targets?

PR #384, opened 2018-05-23
libc default is 'unknown' unless linux target

mark-buer:
For #383

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -280,6 +280,7 @@ module.exports.evaluate = function(package_json,options,napi_build_version) {
     var v = package_json.version;
     var module_version = semver.parse(v);
     var runtime = options.runtime || get_process_runtime(process.versions);
+    var platform = options.target_platform || process.platform;
     var opts = {
         name: package_json.name,
         configuration: Boolean(options.debug) ? 'Debug' : 'Release',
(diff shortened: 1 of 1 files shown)


nicolasnoble:
So, I'm not sure about your suggestion, @mark-buer. If you are cross-compiling, you should know what to put as options when invoking `npm rebuild`. IMHO, I would rather have the user of the cross-compilation tool to pass the `--target_libc` option. For instance, even if you are defaulting to glibc because you are cross-compiling to alpine, you want to override this with `musl`, instead of `glibc` or `unknown`.

Sure, `node-pre-gyp` could have some sort of logic to default to `unknown` as you are suggesting in this PR, but it's kind of muddying the waters even further, and will create even more weird bugs in the case someone cross-compiles for Alpine for example.

My stance would be a tiny stricter than this to be honest: if cross compiling, (which I'm not sure how to detect this, but still), `node-pre-gyp` should refuse to work if all of the variables aren't set.

mark-buer:
Fair enough. My fix was a knee-jerk reaction to my encountering a dependencies attempt to download a binary with "win32" _and_ "glibc" present in the same name!

> My stance would be a tiny stricter than this to be honest: if cross compiling, (which I'm not sure how to detect this, but still), node-pre-gyp should refuse to work if all of the variables aren't set.

I think I agree with the sentiment of your stance. One is either autodetecting everything and can specify/override nothing, or, one is cross compiling and must specify everything.

But I'd still argue that `--target_libc`, auto detected or not, makes no sense for some targets (such as win32 which was the original problem I was trying to solve). The ONLY acceptable target_libc option for win32 is unknown. The user shouldn't be able to specify an illegal combination of variables.

How about - if cross compiling, node-pre-gyp should refuse to work if all of the _relevant_ variables aren't set (for that family)?

nicolasnoble:
Yeah, that's reasonable too. I'm still not sure how node-pre-gyp could detect cross compilation to be honest :-)

In all cases, if you look at https://github.com/electron-userland/electron-builder/issues/3049, I'm arguing that since electron-builder kinda knows about node-pre-gyp, it should also be responsible for setting this variable properly.

springmeyer:
> I'm arguing that since electron-builder kinda knows about node-pre-gyp, it should also be responsible for setting this variable properly.

👍  hence going to close this. thanks all.

Issue #402, opened 2018-07-02
Improve N-API support


jschlight:
I'd like to create a PR to improve support for N-API builds. I'm posting my proposed changes here for review and comment by the community. 

## Support NAN and N-API builds from the same project

For some projects, it is desirable to build both legacy NAN modules and N-API modules from the same package. In the current node-pre-gyp implementation, if the project specifies one or more N-API builds by including the `binary.napi_versions` property in the `package.json` file, and the Node version being built against does not support N-API, node-pre-gyp issues the following error message:

```
node-pre-gyp ERR! stack Error: The N-API version of this Node instance is undefined. This module supports N-API version(s) 1,3. This Node instance cannot run this module.
```

The `package.json` substitution string `{node_abi_napi}` is intended facilitate NAN and N-API builds from the same project. It can be used to distinguish projects that need this capability. 

I propose to modify the current build behavior to attempt a NAN build when all of the following conditions are true:

- The project requests one or more N-API builds by including the `binary.napi_versions` property in the `package.json` file.
- The version of Node being built against does not support a sufficient N-API version to create a successful N-API build.
- The `binary.package_name` property in the `package.json` file contains the text `{node_abi_napi}{napi_build_version}`.

The code that attempts downloads would follow the same rules. 

For new modules that are designed for N-API only, there is no need to include the `{node_abi_napi}{napi_build_version}` text. They would not trigger this special rule. 

The documentation would also be updated to describe this pattern and its usage. 

The C/C++ code would need to determine whether the build being requested is for NAN or N-API. The `NAPI_VERSION` symbol communicated to the compiler by node-pre-gyp can be used for this purpose. 

## Verify each napi_build_version against what’s supported by Node

When building for publishing, node-pre-gyp will verify that the build for each N-API version specified in the package.json file is supported by the version of Node being used for the build. If the version is not supported, node-pre-gyp will issue a warning. If no version can be built, node-pre-gyp will issue an error. This will permit add-ons to be tested against older versions of Node that do not support newer N-API versions.

## Implement a `--build-latest-napi-version-only` argument

Implement a new build argument, `--build-latest-napi-version-only`, that will build only the highest supported N-API version. This will be useful for test builds where only the most current N-API build is of interest by suppressing builds that are eventually ignored.

## Improve recovery from build errors

Improve recovery from build errors so that a configure is not required after a build failure.

## Documentation update

[...]

PR #405, opened 2018-07-10
N-API Improvements

jschlight:
This pull request implements a number of improvements to the node-pre-gyp N-API support for issue https://github.com/mapbox/node-pre-gyp/issues/402.

### N-API builds for only those N-API versions supported by the Node instance

node-pre-gyp now verifies the N-API build versions being requested and builds only those versions supported by current Node instance. If no version can be built, an error message is generated. 

### New `--build-latest-napi-version-only` build argument

A new `--build-latest-napi-version-only` build argument instructs node-pre-gyp to build only the highest N-API version supported by the Node instance. This is useful in CI and test environments where only the most current N-API build is needed. 

### Support for NAN and N-API builds from the same project

This PR introduces the ability to build both N-API and NAN binaries from the same package. 

Placing the configuration value `{node_napi_label}` in the package.json `binary.package_name` property instructs node-pre-gyp to build all viable N-API binaries supported by the current Node instance. If the current Node instance does not support N-API, node-pre-gyp will request a traditional, non-N-API build. 

The configuration value `{node_napi_label}` is set by node-pre-gyp to the type of build created. For N-API builds, the string contains the N-API version with values like `napi-v3`. For traditional, non-N-API builds, the string contains the ABI version with values like `node-v46`.

The C/C++ symbol `NAPI_VERSION` can be used to distinguish N-API and non-N-API builds. 

There are more details in the updated main `README.md` file.

### Build support for Node 4

This PR fixes a JavaScript language issue that was preventing builds against some versions of Node 4. 

### Better recovery from build errors

A `configure` is no longer required in the case where a `build` fails. 

### `NAPI_VERSION` replaces `NAPI_BUILD_VERSION`

Previously, the documentation recommended defining the C/C++ symbol `NAPI_BUILD_VERSION`. The documentation now recommends the symbol `NAPI_VERSION` which is used by the N-API C/C++ headers to define the API for that specific N-API version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,23 +319,25 @@ This value is of importance in two areas:
 1. The C/C++ code which needs to know against which N-API version it should compile.
 2. `node-pre-gyp` itself which must assign appropriate path and file names to avoid collisions.
 
-### Defining `NAPI_BUILD_VERSION` for the C/C++ code
+### Defining `NAPI_VERSION` for the C/C++ code
 
 The `napi_build_version` value is communicated to the C/C++ code by adding this code to the `binding.gyp` file:
 
 ```
 "defines": [
-    "NAPI_BUILD_VERSION=<(napi_build_version)",
+    "NAPI_VERSION=<(napi_build_version)",
 ]
 ```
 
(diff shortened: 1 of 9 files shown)


springmeyer:
Thank you @jschlight

Issue #391, opened 2018-06-03
it builds from source twice when server hosting binaries responds with 403


x-yuri:
[`install()`][1] function eventually calls [`place_binary()`][2]. The latter attaches both [`close`][3] and [`response`][4] event handlers to request. In case of 403 response, [`after_place()`][5] is called twice. As [is][6] [`do_build()`][7]. As such, we get two builds performed in parallel, which generally fail.

Steps to reproduce:
```
yarn add node-zopfli@2.0.2
```

[1]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L159
[2]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L204
[3]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L102
[4]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L108
[5]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L208
[6]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L211
[7]: https://github.com/mapbox/node-pre-gyp/blob/v0.10.0/lib/install.js#L129

kevinmehall:
It seems this can also occur on a 404. For example https://gitlab.com/drom/jtag/-/jobs/80286481:
```
node-pre-gyp WARN Using request for node-pre-gyp https download 
node-pre-gyp WARN Tried to download(404): https://github.com/tessel/node-usb/releases/download/1.3.2/usb_bindings-v1.3.2-node-v64-linux-x64.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for usb@1.3.2 and node@10.6.0 (node-v64 ABI, glibc) (falling back to source compile with node-gyp) 
node-pre-gyp WARN Pre-built binaries not installable for usb@1.3.2 and node@10.6.0 (node-v64 ABI, glibc) (falling back to source compile with node-gyp) 
node-pre-gyp WARN Hit error Connection closed while downloading tarball file 
make: Entering directory '/builds/drom/jtag/node_modules/usb/build'
  CC(target) Release/obj.target/libusb/libusb/libusb/core.o
make: Entering directory '/builds/drom/jtag/node_modules/usb/build'
  CC(target) Release/obj.target/libusb/libusb/libusb/core.o
```

[...]

PR #408, opened 2018-07-26
Fix issue with `close` event always being emitted

addaleax:
This addresses a changed behaviour in Node 10, resulting
from the fact that the `close` event is always being emitted
now. This would have lead to the `place_binary()` callback being
called twice when the HTTP request failed (with e.g. a 404).

Refs: https://github.com/nodejs/node/issues/21063
Fixes: https://github.com/mapbox/node-pre-gyp/issues/391

Edit: The original breakage was reverted in Node 10.8.0 since it was an unintentional semver-major change, but *will* be kept for Node 11+.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -92,6 +92,7 @@ function place_binary(from,to,opts,callback) {
         if (!req) return callback(new Error("empty req"));
         var badDownload = false;
         var extractCount = 0;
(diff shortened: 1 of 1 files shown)


springmeyer:
Thank you. Will tag a new release with this in a moment: v0.10.3

Issue #415, opened 2018-08-31
The node-pre-gyp env variable PACKAGE_NAME_binary_host_mirror makes hosting requisites of multiple packages in a single artifactory difficult


ilatypov:
I found that bcrypt, up to and including the latest version 3.0.0, will erase the URL path in the `bcrypt_lib_binary_host_mirror` variable because its `remote_path` started with a forward slash.  

It would be nice to avoid this.

Also, having a single `binary_host_mirror` variable pointing to a root for binaries of different packages would simplify the use of a local artifactory.

PR #416, opened 2018-08-31
Simplify use of binaries off a local artifactory

ilatypov:
Allow installing a package with many dependent packages using node-pre-gyp binaries off the same local artifactory.

Avoid erasing the subpath in the mirror value.

Corrects #415 .


springmeyer:
sorry nobody reviewed this promptly @ilatypov - since enough time has passed and the code in master has changed, I'd like to ask you to re-create this PR and provide a new one if this is still needed. thanks.

PR #432, opened 2018-11-26
fix node gyp from electron-spawn

kewde:


kewde:
**builds with unparsed options**
> error MSB4057: The target "/p:FOO=bar" does not exist in the project

kewde:
@springmeyer 
I've had to remove two tests - the ability to pass unparsed options to `node-gyp` is broken across versions for _windows_.
A patch has been submitted to https://github.com/nodejs/node-gyp/pull/1616 but it unlikely that this will be merged anytime soon.

I need this fix for `node-sqlite3`, it's not detecting the electron versions properly.

springmeyer:
@kewde okay, thanks for this work.

> I've had to remove two tests - the ability to pass unparsed options to node-gyp is broken across versions for windows.

How about, instead of removing, making them run only for non-windows platforms and adding a code comment that the test could be re-enabled if/after nodejs/node-gyp#1616 is merged?

Also @kewde @nicolasnoble is recently taking on publishing node-pre-gyp releases, so feel free to ping them for review here when this is ready to publish.

lib/util/versioning.js
```diff
@@ -80,7 +80,14 @@ function get_runtime_abi(runtime, target_version) {
     if (runtime === 'node-webkit') {
         return get_node_webkit_abi(runtime, target_version || process.versions['node-webkit']);
     } else if (runtime === 'electron') {
-        return get_electron_abi(runtime, target_version || process.versions.electron);
+        var electron_version = target_version || process.versions.electron;
+        if (!electron_version) {
```
nicolasnoble: It leaves me fairly sad that this is the only solution we have, since electron is being fairly obtuse about its own versions here. I'm wondering how other similar projects are doing this.

I'd love to see a comment here that explains the situation of this fallback however. Something along the lines of

```
// If we don't have an electron version, this means we're in a fork.
// We are then forced to do some heuristics to determine electron's version.
// See https://github.com/electron/electron/issues/9058 for details.
```

Maybe also we could do a node-pre-gyp environment variable override as well, in case the heuristic fails, as a last resort for users? Something along the lines of `process.env.NODE_PRE_GYP_OVERRIDE_ELECTRON_VERSION` maybe?

lib/util/versioning.js
```diff
@@ -80,7 +80,14 @@ function get_runtime_abi(runtime, target_version) {
...
+            // TODO PR something to electron to pass in the version number for forks
+            // https://github.com/electron/electron/issues/9058
+            try { electron_version = require('electron/package.json').version; }
```
murgatroid99: If electron is installed as a global package, this will not work because `require` does not access global packages.

lib/util/versioning.js
kewde: I've briefly looked at the C++ source code of electron and it seems to me like environment variable ELECTRON_RUN_AS_NODE can be overloaded with the version number as it only checks whether it has been defined or not. I would have to verify this though.

It remains a strange and suboptimal solution though.

nicolasnoble: I was more talking about the case where something goes just wrong, and as a last resort, the end user could always use some mechanism to override the automatic detection. Not trying to inject the dependency through some kludge, but rather have an explicit, external, user-controlled override.

springmeyer:
this went stale. if others still see fixes needed please recreate a new PR against latest node-pre-gyp master branch.

Issue #458, opened 2019-05-17
Node-pre-gyp doesn't work in Yarn “offline mode”


murtadha:
In our testing and deployment environment, we use Yarn in [offline mode](https://yarnpkg.com/blog/2016/11/24/offline-mirror/). This makes it impossible to use any NPM packages that use `node-pre-gyp` because it doesn't have a way to provide an alternative for downloading binaries if an internet connection isn't available. See this [example repo](https://github.com/rckclmbr/node_pre_gyp_offline) that shows a very simple way to reproduce this failure. 

One alternative we can consider is to allow `node-pre-gyp` to accept a path that points to where offline binaries are stored. Has anyone explored this before or run into a similar issue?

I will send a PR to demonstrate what we are looking for

murtadha:
Hey @nicolasnoble, when do you plan to release this? We are waiting to pull this fix into our repo. Thank you!! :)

PR #459, opened 2019-05-17
Support extracting compiled binary from local offline mirror

murtadha:
See #458 

This makes it possible to pass in a path to a local file using the `{module_name}_binary_host_mirror` command line argument (or env variable if using Yarn). The `file://` prefix is used to determine that this is pointing to a local offline mirror, instead of a URL.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -91,11 +91,9 @@ function place_binary(from,to,opts,callback) {
         if (err) return callback(err);
         if (!req) return callback(new Error("empty req"));
         var badDownload = false;
-        var extractCount = 0;
         var hasResponse = false;
-        var tar = require('tar');
 
-        function afterTarball(err) {
+        function afterExtract(err, extractCount) {
             if (err) return callback(err);
             if (badDownload) return callback(new Error("bad download"));
             if (extractCount === 0) {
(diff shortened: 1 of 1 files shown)


lib/install.js
```diff
@@ -235,7 +260,12 @@ function install(gyp, argv, callback) {
                     if (err) {
                         after_place(err);
                     } else {
-                        place_binary(from,to,opts,after_place);
+                        var fileName = from.startsWith('file://') && from.replace(/^file:\/\//, '');
+                        if (fileName) {
+                            extract_from_local(fileName, to, callback);
```
nicolasnoble: You're passing `callback` here, but I think you meant `after_place`. Right now, if the `extract_from_local` function fails, (and it can, with file not found for instance), it'll not potentially fallback to building from source, which is the role of `after_place`.

murtadha:
@nicolasnoble, thanks for the quick review. I made the changes, update the commit, and force-pushed, but not sure this is the preferred way to contribute here.

Would you prefer I make a commit on top of it to fix this one line instead?

nicolasnoble:
I'm fine either way :-)

nicolasnoble:
I'll publish this as 0.13.1 shortly.

murtadha:
Than's great, thanks so much @nicolasnoble! When do you expect it to be published? Looking to pull this into our repo soon so we can fix this longstanding issue :)

Thanks for the quick turnaround!!

murtadha:
Hey @nicolasnoble, when are you planning to publish the new release? We are waiting on this to unblock issues we are facing with our environment :)

murtadha:
pinging this @nicolasnoble

mrsimb:
will this be merged?
i need a way to get prebuilt binary locally due to corporate proxy issues.

PR #483, opened 2019-10-25
Update ABI crosswalk with new information

murgatroid99:


murgatroid99:
I added #434 to the changelog assuming it will be merged.

PR #489, opened 2020-02-20
Pin mkdirp version to 0.5.1

apple502j:
mkdirp 1.0.0 introduced a massive breaking change, so pin to 0.5.1 until everything is fixed. This would fix errors when installing.


raksooo:
The current dependency is on `^0.5.1` which means it will always stay on 0.x.x.

PR #491, opened 2020-03-06
update s3 instructions

contra:
Instructions were out of date - this updates the policy included to:

- Add HeadBucket which is needed to publish (pre-gyp fails without this permission)
- Properly scope ListBucket to a bucket ID, otherwise the resource is ignored and ListBucket is able to list any bucket (this is shown as a warning in the AWS UI)

I think the instructions could be a lot simpler and instruct the user to paste the JSON into the JSON editor, then replace the bucket ID - this makes it more foolproof/secure. I'll leave that out for now, can do it in a future PR if the maintainers agree.


springmeyer:
contra - thanks for offering this. looks nearly ready to merge just needs updated against latest master. can you do that?

contra:
@springmeyer Resolved

springmeyer:
Thanks @contra - CI failures are unrelated to your PR (sorry about that). @bmacnaughton - are these CI failures what you would expect for an external PR? Is there any way to avoid them?

bmacnaughton:
@springmeyer. yes, i the errors are what i'd expect to fail on an external PR. because travis doesn't provide the AWS credentials to jobs initiated by external contributors it fails in that case. (the tests are those that use verify the proxy access works and those are not compatible with mocking s3, so they require real credentials).

the only way to avoid them that i have thought of is to not run the tests if the credentials aren't present. that might be able to be combined with an additional check based on who is submitting the PR.

springmeyer:
> the only way to avoid them that i have thought of is to not run the tests if the credentials aren't present

👍 this seems like the easiest solution. Could you put up a PR for this @bmacnaughton?

README.md
```diff
@@ -433,28 +433,37 @@ First, get setup locally and test the workflow:
...
+            "Sid": "objects",
+            "Effect": "Allow",
+            "Action": [
+                "s3:PutObject",
+                "s3:GetObjectAcl",
+                "s3:GetObject",
+                "s3:DeleteObject",
+                "s3:PutObjectAcl"
+            ],
+            "Resource": "arn:aws:s3:::node-pre-gyp-tests/*"
```
springmeyer, line 452:
```suggestion
            "Resource": "arn:aws:s3:::your-bucket-name/*"
```

README.md
```diff
@@ -433,28 +433,37 @@ First, get setup locally and test the workflow:
...
+        },
+        {
+            "Sid": "bucket",
+            "Effect": "Allow",
+            "Action": "s3:ListBucket",
+            "Resource": "arn:aws:s3:::node-pre-gyp-tests"
```
springmeyer, line 458:
```suggestion
            "Resource": "arn:aws:s3:::your-bucket-name"
```

bmacnaughton:
i will get to it by the end of the weekend at the latest.

PR #498, opened 2020-05-04
feat: Add GitHub Action to PR ABI changes

nschonni:
Runs nightly and creates a PR

### DIFF
--- a/.github/workflows/nightly-crosswalk.yml
+++ b/.github/workflows/nightly-crosswalk.yml
@@ -1,0 +1,28 @@
+name: Nightly Sync of ABI Crosswalk
+
+on:
+  schedule:
+  - cron: '0 3 * * *' # run daily at 3am
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


nschonni:
Alternately, this could just push directly to master for the changes

springmeyer:
@nschonni thanks for starting this. My take is that this could be useful if it also auto-created tags. Without that we still need a manual step involved.

So, curious if we could step back a bit: can you add more details of what problem you think this overall needs to solve? Last I recall the abi checking worked except for when new major versions are released.

nschonni:
This is from almost a year ago, so I'm not really sure of the details anymore. I would guess that it was around the CI being perma-red at the time and the project mostly inactive. This was just the same approach from https://github.com/chicoxyzzy/node-releases that automated similar data to allow things to stay in sync without manual interventions by the maintainers

springmeyer:
Okay, thanks @nschonni - we're green again so hopefully this will not be needed. Please don't hesitate to create another issue/PR if you find it is needed again.

nschonni:
This might still be useful as an automation for the project so that when ABI versions change or new versions are released, you get the PR with the updates without having to remember.
I think the whole tagging/release would probably be a separate task

PR #504, opened 2020-07-08
Respect npm config runtime, platform, arch

lucasyvas:
We are currently building our Windows platform on Linux. `node-pre-gyp` does not seem to support many ways I could find to influence these settings outside of command options. This PR aims to respect `.npmrc` config much like `prebuild-install` does here:

https://github.com/prebuild/prebuild-install/blob/406b0513833be7c9e909128e642a42e82242f720/rc.js#L35

If there is a way to do this through an environment variable or something else, please advise and I can close this.

The particular aim of these particular options is to support installing a prebuilt electron module for `sqlite3` on a different platform and architecture.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -279,7 +279,7 @@ module.exports.evaluate = function(package_json,options,napi_build_version) {
     validate_config(package_json,options); // options is a suitable substitute for opts in this case
     var v = package_json.version;
     var module_version = semver.parse(v);
-    var runtime = options.runtime || get_process_runtime(process.versions);
+    var runtime = options.runtime || process.env.npm_config_runtime || get_process_runtime(process.versions);
     var opts = {
         name: package_json.name,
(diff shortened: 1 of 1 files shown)


springmeyer:
@lucasyvas thanks for providing this. I'm open to accepting this if:

 - you update to latest `master`
 - you clearly explain why this is needed
 - you add tests

lucasyvas:
@springmeyer I don't personally have a use anymore. For context, I was working with a team that built a Windows app within WINE and not having the platform overrides for prebuilt binaries was a pain.

If you don't see any value in this use case yourself, close it!

springmeyer:
Okay, thanks @lucasyvas - I can see value to being able to pass in options in a way that is not from the command line. But at the same time the command line options should work for most (all?) usecases, so I'm inclined to keep things simple and only support the command line for now until others raise the need as critical.

PR #511, opened 2020-08-05
feat: use an agent to set proxy when using needle

bmeriaux:
because needle does not make a CONNECT request when needed, it simply forward request which is not correct for https request, so we need to follow: https://github.com/tomas/needle#more-advanced-proxy-support


bmeriaux:
@springmeyer could you look at this pr, please , we really need to fix this problem behind a proxy :/

springmeyer:
@bmeriaux this looks good. I don't have a way to test this myself, so two questions:

 - Can you think of a way to test this so that it does not regress in the future?
 - You can confirm it works for your usecase?

Also, I presume you need this since you don't want to use request instead?

bmeriaux:
I try to add a test to check the presence of an agent.
I did a real life test with the fix and its worked :)

lekoaf:
@bmeriaux @springmeyer  How's it going with this PR? This is something me and my team would need for a project we're working on. Thanks! :smiley:

springmeyer:
Looks like the next step with this PR is to get the test working on windows. Looks broken currently due to:

```
# should set a agent for proxy with needle
C:\projects\node-pre-gyp\test\install-proxy.test.js:39
  process.chdir('test/app1');
          ^
Error: ENOENT: no such file or directory, uv_chdir
    at Error (native)
    at Test.<anonymous> (C:\projects\node-pre-gyp\test\install-proxy.test.js:39:11)
    at Test.bound [as _cb] (C:\projects\node-pre-gyp\node_modules\tape\lib\test.js:80:32)
    at Test.run (C:\projects\node-pre-gyp\node_modules\tape\lib\test.js:96:10)
    at Test.bound [as run] (C:\projects\node-pre-gyp\node_modules\tape\lib\test.js:80:32)
    at Immediate.next (C:\projects\node-pre-gyp\node_modules\tape\lib\results.js:83:19)
    at runCallback (timers.js:672:20)
    at tryOnImmediate (timers.js:645:5)
    at processImmediate [as _immediateCallback] (timers.js:617:5)
npm ERR! Test failed.  See above for more details.
```

springmeyer:
made obsoleted by #551

PR #519, opened 2020-10-24
fix: if passed python option set environment variable

ewanharris:
Previously if passed the --python option node-gyp would always use it, now the only way to force
using that value is to set an environment variable pointing to it

This gets a little closer to getting the tests running, locally I'm still seeing some issues in tests failing to compile due to changes in Node that impact the c++ code, unfortunately those fixes are beyond me 😄


springmeyer:
Thanks for trying to fix this. A while back (apologies I don't have the link handy) I fixed this by adjusting the test instead.

PR #528, opened 2020-12-04
Update abi_crosswalk.json

Pomax:
A number of releases [have occurred](https://nodejs.org/en/download/releases/) since the last update.


Pomax:
It also looks like something went rather badly wrong with the CI matrix, it seems to only test Node 6 and Node 10... 

image

springmeyer:
Thanks, applied in https://github.com/mapbox/node-pre-gyp/commit/834bbe0353f627a4cecfadd30cae946b3ca49f82

PR #529, opened 2020-12-09
enable differentiating versions of linux in naming template

bmacnaughton:
If a node package includes pre-built libraries then in some cases there can be linux-distribution-specific libraries. The alpine distribution is the most notable case because it uses `musl` while most distributions use `glibc`. This PR implements the `--distinguish_linux` option to allow inserting a specific linux ID in templates via the variable `linux_name`.

In the following, `id` is the value of the `id` property in `/etc/os-release` (or variant of the same).

`--distinguish_linux` by itself sets `linux_name` to `id`
`--distinguish_linux=alpine` sets `linux_name` to 'alpine' only if `id` is 'alpine'.
`--distinguish_linux=alpine,centos` sets `linux_name` to `id` only if `id` is 'alpine' or 'centos'

`linux_name` is just 'linux' by default.

The tests are not passing but they are not passing with HEAD or with the latest release, so I wanted to get this PR in to solicit feedback in the meantime. I am happy to change the option name or any other items that you suggest.

I will take a look at the tests as time allows; it looks like I need to create an environment with clang and some other bits that I don't normally use. Any advice on test setup is welcome.

I did update abi_crosswalk.json but didn't do the other items in the contribution guidelines. I will work on tests but the other items looked more like a checklist for you because they included `npm publish` so I wanted to check with you before doing others.

**update**

since the initial PR I added a feature that allows toggling between staging and production s3 buckets when executing commands. the update is described in the readme.

i have also updated all the tests in .travis.yml to use nan in order to make them more node-version independent.
the webkit test required some changes as the package location and naming changed.

oh, and i probably overstepped as i took out all the jshint checks (in favor of using a globally installed eslint while editing). it's easy enough to put back in if that's a major issue.


springmeyer:
@bmacnaughton this PR has many great improvements - thanks for offering it up. I just did some modernization of the code in master and made several similar changes (like switching to eslint and fixing the webkit tests). Sorry for not seeing this first otherwise I would have wanted to start from your work. But I'm just seeing this now. So, seems like the best next step would be for you to re-integrate your changes against latest master, if you can. Ideally separate them out into different PRs for different types of cleanup and improvements.

bmacnaughton:
i will take a look and see what it entails. i probably should have referenced you in the original PR.

bmacnaughton:
@springmeyer - one more question: you're ok with changing `var` to `const` or `let` or is deep backwards compatibility required?

bmacnaughton:
ok, i've submitted two separate PRs. i have not tested either but will try to get to that tomorrow.

let me know if there is work yet-to-be-done on the test changes; i see you've modified them.

springmeyer:
Yes, I'm 👍  on using `const` and getting rid of `NAN`. I've done both already in master as I think you've found. I'll review your PRs shortly. Thank you.

PR #532, opened 2021-01-02
enable differentiating versions of linux in naming template (solo)

bmacnaughton:
If a node package includes pre-built libraries then in some cases there might be linux-distribution-specific libraries. The alpine distribution is the most notable case because it uses musl while most distributions use glibc. This PR implements the `--distinguish_linux` option to allow inserting a specific linux ID in templates via the variable linux_name.

In the following, `id` is the value of the `id` property in `/etc/os-release` (or variant of the same).

`--distinguish_linux` by itself sets `linux_name` to `id`
`--distinguish_linux=alpine` sets `linux_name` to 'alpine' only if `id` is 'alpine'.
`--distinguish_linux=alpine,centos` sets `linux_name` to `id` only if `id` is 'alpine' or 'centos'

`linux_name` is 'linux' by default.

this also fixes the checks for required binary properties at [versioning.js 211](https://github.com/mapbox/node-pre-gyp/compare/master...bmacnaughton:linux-differentiation-only?expand=1#diff-731dda4952e8957a5d9387cd5ed656a10a7a6fd6e22b1ab3ef12502a9917e03dR211)


springmeyer (commented):
This is a useful addition thanks. My only concern/question is: why not have this available as a variable by default on linux? Why require the `--distinguish_linux` to be passed?

springmeyer:
Two other thoughts:

 - Ping me when the tests are passing (and probably best to pull in latest master before you try to get them passing)
 - The arbitrary `--toolset` option is what I've been using to account for linux specific differences like this. I've also used it to package different type of linux binaries like those with and without address sanitizer. You might consider trying it out for handling alpine musl/glibc issues. If not and you want to make the os-release available, thats just fine. I only wanted to explain why I'd not yet added something like this: I've been getting by with `--toolset`.

bmacnaughton:
> This is a useful addition thanks. My only concern/question is: why not have this available as a variable by default on linux? Why require the --distinguish_linux to be passed?

the main reason is that i only care about alpine. if there is no option and it always sets `linux_name` to whatever is in `/etc/os-release` then it's not possible to have only two linux variations, i.e., alpine and linux. each variant of linux would then need to be built for the no-option case.

> toolset

it looks like could solve the issue if i embed the logic i added in this PR to the script line. but i don't (after a quick look) see how i would load a different binary based on the linux variant otherwise. so the package that invokes node-pre-gyp would need to execute a script that performed the new logic and pass the results to the `--toolset` option. is this the way you're thinking about it?

bmacnaughton:
@springmeyer - tests are passing. i can rebase on master if you'd like but i don't believe there are any merge conflicts.

springmeyer:
> it looks like this could solve the issue if i embed the logic i added in this PR to the script line

Correct. Say you have two different docker builds for alpine and normal linux. You'd add `/{toolset}/` into your versioning string in package.json, then in the alpine build you'd publish with `node-pre-gyp publish --toolset="alpine"` and with the normal linux build you'd just do `node-pre-gyp publish`. In the latter case the `toolset` would not be set and would be collapsed away. Then when you want to go install each version you'd do either `npm install` or `npm install --toolset=alpine`.

bmacnaughton:
given that solution i'd prefer that the code is in node-pre-gyp if that's ok with you. the reason is that i would prefer not to have to ask the end user to `npm install` or `npm install --toolset=alpine`. if the code is in node-pre-gyp then that's transparent.

springmeyer:
Sure, not having to ask/require the end user seems useful.

springmeyer:
Let's get your other PR landed first, then rebase this one, then this one can land. Sound good?

bmacnaughton:
Sounds great. I appreciate your working with me on these PRs.

bmacnaughton:
i have rebased this PR and updated the code to use let and const now. in the meantime, i've wondered if it might make sense to just set `platform` to the linux name as opposed to creating the new configuration value, `linux_name`. what do you think?

i'm going to work on some tests in the meantime.

bmacnaughton:
i have rebased this PR to include the --directory option fix.

in the meantime, i've wondered if it might make sense to just set `platform` to the linux name as opposed to creating the new configuration value, `linux_name`. i don't think it will break anyone because they have to use the option to get that change. 

@springmeyer what do you think? if you agree i'll modify this as well as add a test or two.

springmeyer:
> in the meantime, i've wondered if it might make sense to just set platform to the linux name as opposed to creating the new configuration value, linux_name. 

Interesting idea. Also, I'm now remembering that node-pre-gyp already supports `{libc}` which was intended to allow versioning for alpine: https://github.com/mapbox/node-pre-gyp/pull/326. Did you see this? Do you think you could accomplish what you need with the `{libc}` variable?

bmacnaughton:
no, libc won't do it because we distribute a separate library that has an alpine-specific build. so our alpine build includes a different library in the tar file which is linked using the $ORIGIN rpath.

bmacnaughton:
this does what is required for me. i'm going to close this PR because my need for it is addressed by using `{libc}`. feel free to reopen it if you can think of other uses for this.

PR #537, opened 2021-01-07
fix to read correct package.json when --directory option is used

bmacnaughton:
My change to consolidate reading package.json in the `Run` constructor failed to account for the `--directory` option. fixing It required some rearrangements because it as a little bit of chicken-and-egg - argv needed to be parsed to know that `--directory` was specified but `parseArgv` required access to package.json.

This set of changes fixes that by moving the calls to `parseArgv()` into the `Run` constructor. This requires that `argv` is passed as an argument whenever `new Run()` is invoked. And package.json is now read in `parseArgv()` instead of directly in the `Run` constructor. The change results in one less read of package.json.

For testing, the tests that invoked `new Run()` were moved out of `versioning.test.js` and into a new file, `run.test.js`. Tests to verify that the `--directory` option works as expected were added to that. Double checking that testing would be appreciated.


bmacnaughton:
if i am interpreting the coverage changes correctly the reduction in coverage is not due to any of these changes; it seems to be the same issue as in the previous PR.

springmeyer:
Looked at the coverage and just realized what was going on before and I missed understanding. It is that some tests are conditionally run: https://github.com/mapbox/node-pre-gyp/blob/a55d5fce623fa6977deaee82e70a5624c8ee2268/test/build.test.js#L274-L338. And I gather that travis is not going to set those secure variables for external PRs. Any chance you have experience with mocking S3 calls and could work on that? Ideally these tests could be run without needing S3 credentials to avoid this PR coverage drop problem...

bmacnaughton:
interesting; i hadn't thought about it but it makes sense. if they did supply the secrets then i could capture and report them through some mechanism.

i'll take a look at mocking solutions; i haven't tried them before but there are a few out there though they haven't been updated recently.

springmeyer:
> i'll take a look at mocking solutions; i haven't tried them before but there are a few out there though they haven't been updated recently.

One external contributor (@allenluce) added some mocking in https://github.com/mapbox/node-pre-gyp/blob/master/test/fetch.test.js. That looks to support http gets at least, so it could be used in the `test/build.test.js`. The question is how to handle the s3 PUTs...

bmacnaughton:
i'm going to spend a little time on this: https://github.com/MathieuLoutre/mock-aws-s3

the nock responder could use the s3 mocks to fetch the desired objects.

nock - the README only states support up to node 9 but their CI matrix does not 10, 12, 14.  so it's probably good. i'll work on this today.

bmacnaughton:
i think it makes sense for the mock changes to be a completely different PR. the changes are really independent of this PR and the linux-differentiation PR. what do you think of getting this PR and the linux-differentiation PR out of the way while i work on this as a separate PR?

bmacnaughton:
so far, the only way i see to allow mocking for testing is to change both released code in addition to test code. the released code will need to be able to load the mock-s3-code if an environment variable is present (`node_pre_gyp_mock_s3`?)

in addition to the tests, `lib/info.js`, `lib/publish.js`, `lib/unpublish.js`, and `scripts/clear_bucket.js` would need to honor that env var.

springmeyer:
👍  to get this in separate from mocking. 👍  to change code in lib/ to get mocking working. Thanks for looking into it! I'll merge this now.

springmeyer (approved):

PR #539, opened 2021-01-09
change tests to use mock s3

bmacnaughton:
if the env var `node_pre_gyp_mock_s3` is not empty then it is used as the root directory for the mocked s3. e.g., if set to `/tmp/mock` then the `mapbox-node-pre-gyp-public-testing-bucket` will be directory `/tmp/mock/mapbox-node-pre-gyp-public-testing-bucket` and each key will be written to that bucket.

It is now necessary to call `s3_setup.get_s3()` instead of directly instantiating `new AWS.S3()`. only `s3_setup.js` requires `aws-sdk` now because `s3_setup.get_s3()` uses `process.env.node_pre_gyp_mock_s3` to choose between the real S3 and the mock s3.

because the install command uses https rather than s3 to fetch binaries, http must be mocked in testing for that command. i didn't want to embed a decision about mocking http in `lib/install.js` so it got a bit more complicated.

1. there is a new module `test/create-mock-s3.js` that instantiates `nock` and reads from the `node_pre_gyp_mock_s3` directory.
2. all test apps "install" scripts are now: `"${node_pre_gyp_command:-npx node-pre-gyp} install --fallback-to-build",`
3. `test/run.util.js` was modified so that if mocking s3:
    - if the command is `node-pre-gyp`, then it is prefixed with `node -r create-mock-s3.js`
    - if the command is `npm`, then `node_pre_gyp_command` is defined as `node -r create-mock-s3.js node-pre-gyp` and inserted into the exec'd jobs environment. it becomes the command that `npm install` executes (because of 2).

Miscellaneous changes/notes
- add `--npm install --package-lock-only` after installing request
- ignore app7 build artifacts
- it doesn't appear that the `--fallback-to-build=false` arguments in `build.test.js` have an effect in the exec'd tasks.
- `run.util.js` accepts a debugging option

i can't test against mapbox buckets, so it's probably worth a double-check with `npm run no-mock-test`.

### DIFF
--- a/lib/util/s3_setup.js
+++ b/lib/util/s3_setup.js
@@ -24,3 +24,63 @@ module.exports.detect = function(to, config) {
     }
   }
 };
+
+module.exports.get_s3 = function(config) {
+
+  // if not mocking then setup real s3.
+  if (!process.env.node_pre_gyp_mock_s3) {
+    const AWS = require('aws-sdk');
+
+    AWS.config.update(config);
+    const s3 = new AWS.S3();
+
+    // need to change if additional options need to be specified.
+    return {
+      listObjects(params, callback) {
+        return s3.listObjects(params, callback);
+      },
+      headObject(params, callback) {
+        return s3.headObject(params, callback);
+      },
+      deleteObject(params, callback) {
+        return s3.deleteObject(params, callback);
+      },
+      putObject(params, callback) {
+        return s3.putObject(params, callback);
+      }
+    };
+  }
+
+  // here we're mocking. node_pre_gyp_mock_s3 is the scratch directory
+  // for the mock code.
+  const AWSMock = require('mock-aws-s3');
+
+  AWSMock.config.basePath = process.env.node_pre_gyp_mock_s3;
+
+  const s3 = AWSMock.S3();
+
+  // wrapped callback maker. fs calls return code of ENOENT but AWS.S3 returns
+  // NotFound.
+  const wcb = (fn) => (err, ...args) => {
+    if (err && err.code === 'ENOENT') {
+      err.code = 'NotFound';
+    }
+    return fn(err, ...args);
+  };
+
+  return {
+    listObjects(params, callback) {
+      return s3.listObjects(params, wcb(callback));
+    },
+    headObject(params, callback) {
+      return s3.headObject(params, wcb(callback));
+    },
+    deleteObject(params, callback) {
+      return s3.deleteObject(params, wcb(callback));
+    },
+    putObject(params, callback) {
+      return s3.putObject(params, wcb(callback));
+    }
+  };
+
+};
--- a/test/app1/package.json
+++ b/test/app1/package.json
@@ -17,7 +17,7 @@
         "package_name": "{node_abi}-{platform}-{arch}.tar.gz"
     },
     "scripts": {
-        "install": "node-pre-gyp install --fallback-to-build",
+        "install": "${node_pre_gyp_command:-npx node-pre-gyp} install --fallback-to-build",
         "test": "node index.js"
     }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ npm-debug.log
 .nyc_output
 coverage
 test/*/package-lock.json
+test/app7/build-tmp-napi-*
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
       node_js: 10
       before_install:
         - npm install request
+        - npm install --package-lock-only
       script:
         - npm run lint
         - npm run coverage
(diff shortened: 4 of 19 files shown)


bmacnaughton:
appveyor is failing because i can't set an env var at the beginning of the script line (where it uses `os.tmpdir()`).

i don't have much experience making this kind of thing work on windows. any ideas?

lib/util/s3_setup.js
```diff
@@ -24,3 +24,63 @@ module.exports.detect = function(to, config) {
     }
   }
 };
+
+module.exports.get_s3 = function(config) {
```
springmeyer, line 28: [praise] this centralization is great - been thinking node-pre-gyp should move to https://github.com/aws/aws-sdk-js-v3 soon to address the complaint of many that the aws-sdk v2 is heavy/leads to large package downloads. So this centralization will help with that future move.

test/app1/package.json
```diff
@@ -17,7 +17,7 @@
         "package_name": "{node_abi}-{platform}-{arch}.tar.gz"
     },
     "scripts": {
-        "install": "node-pre-gyp install --fallback-to-build",
+        "install": "${node_pre_gyp_command:-npx node-pre-gyp} install --fallback-to-build",
```
springmeyer, line 20: [question] what this is change doing? It is optional or needed?

springmeyer:
This looks like great progress, thanks for taking this on!

>  appveyor is failing because i can't set an env var at the beginning of the script line (where it uses os.tmpdir()).

No problem, we can worry about that last.

> i don't have much experience making this kind of thing work on windows. any ideas?

I don't have a windows box to test on, but I can probably figure it out. First idea is to not override `test` in the package.json but rather require us to set an `ENV` setting to a) provide a tmpdir in the appveyor.yml or travis.yml and b) to turn on mocking. So instead of doing anything in package.json, just handle all the complexity in the test code.

test/app1/package.json
bmacnaughton: that's no longer there though i replaced it with a `test/tape-helper.js` which works fine on linux. just pushed; we'll see how it works on windows. i'll need to rethink a few other pieces as well but wanted to give it a shot.

maybe i'm trying to hard not to put test-specific changes in `install.js`. in order to mock http without modifying `install.js` the only solution i came up with was to change the command `node-pre-gyp install`  to `node -r create-mock-s3 node-pre-gyp install`. that loads the create-mock-s3 module (in test, so not even distributed via npm) so doesn't require changes to `install.js` or additional code to be distributed. so detecting that environment variable is required. but my most recent commit moves that logic to `test/tape-helper.js` so no linux-specific syntax is in the script section.

bmacnaughton:
>  set an ENV setting to a) provide a tmpdir in the appveyor.yml or travis.yml and b) to turn on mocking. So instead of doing anything in package.json, just handle all the complexity in the test code.

that's the idea - the env var is `node_pre_gyp_mock_s3`. but it was the loading of the http mock in a way that didn't require changes to `install.js` that was the problem the test command change was solving. i think that will all be in the helper after one more commit.

test/app1/package.json
bmacnaughton: i do not know how to get around this in the test apps without requiring changes to `install.js` for mocking http. i'm guessing you're using https because it's more universal than s3 and it doesn't require aws-sdk to be installed, but one solution is just to move to using s3. it'll be a bigger distribution though.

another solution is to just change the test commands to execute what `npm install` would execute directly. it bypasses package.json but tests the same thing, i believe.

the other solution i can think of is to modify `install.js` to load the mocking module if the `node_pre_gyp_mock_s3` environment variable is present. i avoided that in favor of not modifying "production" code, but it's doable and not a big deal.

bmacnaughton: @springmeyer - here's a better idea of the options i see. the first three do not require any changes to `install.js` for https mocking.

1 use what it's doing now but fix it for windows. that requires yet another `tape-helper`-like module to spawn the tests after modifying the command line. the primary advantage is that there is no additional code in `install.js`; the primary negative is that it mucks around with the actual commands, i.e., we're not testing the exact same thing the user is executing.
2 move to using s3 to install. advantage - no need to mock https in install. negative - requires aws-sdk for install, compatibility not likely to be as good as https.
3 just modify the test commands directly, i.e. instead of `build.test.js` executing `npm install --fallback-to-build=false` it just executes what `npm install` actually does, `node-pre-gyp install`. advantage - no changes to package.json, disadvantage - doesn't use package.json.
4 modify `install.js`. i had been trying to avoid this but i think that goal was a bit off. i'd been thinking of this as something that runs in "production" as opposed to just during installation. there is very little reason not to change `install.js` to add mocking when in testing mode.

if you're ok with 4, i'll do that. that means some of what i did was wasted but i think it's the better solution.

bmacnaughton:
this no longer seems relevant. the other s3 testing PR seems much cleaner than all the script changes, helpers, and env setup.

PR #543, opened 2021-01-18
headObject needs Key

bmacnaughton:
fix publish `headObject()` call

- was using Prefix; now using Key

### DIFF
--- a/bin/node-pre-gyp
+++ b/bin/node-pre-gyp
@@ -79,8 +79,8 @@ function run() {
 
   // set binary.host when appropriate. host determines the s3 target bucket.
   const target = prog.setBinaryHostProperty(command.name);
-  if (target) {
-    log.info('setting binary.host to ' + prog.package_json.binary.host);
+  if (target && ['install', 'publish', 'unpublish', 'info'].indexOf(command.name) >= 0) {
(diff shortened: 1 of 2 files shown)


bmacnaughton:
danger of not testing against real s3 now. i'm not sure the best approach. seems like we should have a manually triggered travis/github-actions job that performs a test using the real s3. this wasn't caught because the mock-s3 doesn't check the validity of parameters as the real AWS.S3 api does.

springmeyer:
Thanks for catching. This should set us up for catching this problem on CI in the future (https://github.com/mapbox/node-pre-gyp/pull/544) at least for the merge to master. I just merged it to master so that future PRs should fail until the bug is fixed. Can you try both a) updating this PR against master and b) creating a fresh one to see if it fails without this PR if you create it (my guess is that it won't since you are an external collab that travis won't expose the publishing keys for, but worth a check).

bmacnaughton:
@springmeyer  - thanks for the quick look.

a) i have rebased this PR on your master changes.
b) #545 is a dummy PR that you can delete/close after checking out. it was based on your master without this PRs changes.

springmeyer (approved):

PR #550, opened 2021-01-28
support node-fetch with proxy

bmacnaughton:
this replaces needle and request with node-fetch
- adds proxy support for both http and https destinations (but only http proxy)
- adds testing with builtin proxy server

changes
- changes messages issued by install - the flow changed dramatically
- requires a test bucket to have a suitable binary in it; i'm using one i set up but it should be replaced by one in a mapbox bucket.

this took a lot longer than i expected; https.request with method: 'get' works differently than https.get. it took me quite a bit of time to try that change because i thought that https.get was just syntactic sugar. i didn't dig into what the difference was. but i have tried many different proxy agents and proxy server implementations as a result.

on the second change point - you're welcome to copy the files in the s3 bucket i set up, or create completely different ones. and `app-proxy/package.json` can be changed. adding the napi property to the binary property highlighted that a bit of additional setup was required before calling `install` directly, so i think it's a good thing to include that in at least one test.

an additional thing i noticed is that node-pre-gyp.find() is only prepared to find one `.node` file. while it's not included in `test/app-proxy` my binding.gyp has two targets and uses a pseudo-target ("action_after_build") that just executes "copies" to put two `<(module_path)` where `_package()` picks them up. in order to fetch the second module the code looks like:

```
'use strict';

const binary = require('@mapbox/node-pre-gyp');
const path = require('path');
const pkgpath = path.resolve(path.join(__dirname, './package.json'));
const binding_path = binary.find(pkgpath);
const bindings = require(binding_path);

module.exports = bindings;

// this is hardcoded as node-pre-gyp doesn't know about multiple targets.
const metrics_path = binding_path.replace('apm_bindings.node', 'ao_metrics.node');
module.exports.metrics = require(metrics_path);
```

i could change it to allow a target name to be passed to `find()` as in `binary.find(pkgpath, 'ao_metrics')` and replace the default target name if it is present. not sure how common this is though and it works just fine the way it is now.


bmacnaughton:
@springmeyer - i should have let you know what i was up to here.

package.json
```diff
@@ -21,8 +21,10 @@
   "main": "./lib/node-pre-gyp.js",
   "dependencies": {
     "detect-libc": "^1.0.3",
+    "http-proxy-agent": "^4.0.1",
+    "https-proxy-agent": "^5.0.0",
```
springmeyer, line 25: One of the reasons I recall you mentioning node-fetch as better than needle was size. After adding these extra deps, does it still look good for fetch overall in terms of being nice and small? e.g. are either of these quite big?

springmeyer:
> changes messages issued by install - the flow changed dramatically
Seems fine.

> requires a test bucket to have a suitable binary in it; i'm using one i set up but it should be replaced by one in a mapbox bucket.

would really prefer a test bucket not having to have state. just to make it safe to clear it / create a new one, etc and not have a setup step that could get lost or be mis-understood. So can you think of a way to avoid needing state but still test things okay?

> this took a lot longer than i expected; https.request with method: 'get' works differently than https.get. it took me quite a bit of time to try that change because i thought that https.get was just syntactic sugar. i didn't dig into what the difference was. but i have tried many different proxy agents and proxy server implementations as a result.

thanks for all the efforts!

> an additional thing i noticed is that node-pre-gyp.find() is only prepared to find one .node file. while it's not included in test/app-proxy my binding.gyp has two targets and uses a pseudo-target ("action_after_build") that just executes "copies" to put two <(module_path) where _package() picks them up. in order to fetch the second module the code looks like:

This limitation seems okay to me, at least. Because I've tended to author my package.json `binary` path such that the location is predictable. The `find` was built in order to enable the runtime path to be dynamic which ended up not being used much by downstream modules. So I'm thinking about deprecating/removing the find support unless others really need it. Feels overcomplicated.

bmacnaughton:
> would really prefer a test bucket not having to have state

can do that, it just adds another test requiring writing to a *real* bucket  because i don't think it's a valid test to use nock to simulate s3 and a proxy-server both. i focused on a test that would run but not require write permissions to a real bucket so it can run on anyone's PR.  let me know.

> size
node-fetch 158 KB + https-agent 26.2 KB + http-agent 17.1 KB = 201,3 KB
needle 217 KB

bmacnaughton:
i did verify that the proxy used in this test acts enough like a proxy that curl worked through it. i don't have other proxies to test against but that seems like a pretty good test.


```
$ node
> proxy = require('./test/proxy.util.js')
> proxy.startServer()
[proxy server running at http://localhost:8124]
```

in another terminal:
```
$ export https_proxy=http://127.0.0.1:8124
$ export http_proxy=http://127.0.0.1:8124
$ curl -v https://apm-bruce-production.s3.amazonaws.com/apm_bindings/v11.0.0/apm_bindings-v11.0.0-napi-v4-glibc-x64.tar.gz > t-curl.out
* Uses proxy env variable https_proxy == 'http://127.0.0.1:8124'
*   Trying 127.0.0.1:8124...
* TCP_NODELAY set
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to 127.0.0.1 (127.0.0.1) port 8124 (#0)
* allocate connect buffer!
* Establish HTTP proxy tunnel to apm-bruce-production.s3.amazonaws.com:443
> CONNECT apm-bruce-production.s3.amazonaws.com:443 HTTP/1.1
> Host: apm-bruce-production.s3.amazonaws.com:443
> User-Agent: curl/7.68.0
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 200 OK
< 
* Proxy replied 200 to CONNECT request
* CONNECT phase completed!
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* CONNECT phase completed!
* CONNECT phase completed!
{ [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [91 bytes data]
* TLSv1.2 (IN), TLS handshake, Certificate (11):
{ [2900 bytes data]
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
{ [333 bytes data]
* TLSv1.2 (IN), TLS handshake, Server finished (14):
{ [4 bytes data]
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
} [70 bytes data]
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
* TLSv1.2 (OUT), TLS handshake, Finished (20):
} [16 bytes data]
* TLSv1.2 (IN), TLS handshake, Finished (20):
{ [16 bytes data]
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=US; ST=Washington; L=Seattle; O=Amazon.com, Inc.; CN=*.s3.amazonaws.com
*  start date: Nov  9 00:00:00 2019 GMT
*  expire date: Mar 12 12:00:00 2021 GMT
*  subjectAltName: host "apm-bruce-production.s3.amazonaws.com" matched cert's "*.s3.amazonaws.com"
*  issuer: C=US; O=DigiCert Inc; OU=www.digicert.com; CN=DigiCert Baltimore CA-2 G2
*  SSL certificate verify ok.
} [5 bytes data]
> GET /apm_bindings/v11.0.0/apm_bindings-v11.0.0-napi-v4-glibc-x64.tar.gz HTTP/1.1
> Host: apm-bruce-production.s3.amazonaws.com
> User-Agent: curl/7.68.0
> Accept: */*
> 
{ [5 bytes data]
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< x-amz-id-2: WrGVgXSf4Daj5M2dqgXjYdNrorTivQByC+4XE7bAF7laz0poAt0n1I2iJ0jzB8E1FuXkU3cSPzY=
< x-amz-request-id: 653CF2D7F6FC0148
< Date: Tue, 26 Jan 2021 16:24:14 GMT
< Last-Modified: Wed, 20 Jan 2021 11:43:46 GMT
< ETag: "d7a84e9b85fe167ea6e9472f1d56ef30"
< Accept-Ranges: bytes
< Content-Type: application/octet-stream
< Content-Length: 91201
< Server: AmazonS3
< 
{ [5 bytes data]
100 91201  100 91201    0     0   140k      0 --:--:-- --:--:-- --:--:--  140k
* Connection #0 to host 127.0.0.1 left intact
```
i then untarred the downloaded file to verify that it was correct (and even went so far as to load it after copying over the .so file it requires).

springmeyer:
thanks for the extra details

> can do that, it just adds another test requiring writing to a real bucket because i don't think it's a valid test to use nock to simulate s3 and a proxy-server both. i focused on a test that would run but not require write permissions to a real bucket so it can run on anyone's PR. let me know.

I guess I'm confused then. To restate my preferences:

 - Mocking ideally
 - If that does not work, writing to a real bucket, but not requiring that bucket to have state

I think its okay to have a test that only runs on contributor/not external PRs, as long as you can work with me to ensure it is passing. Happy to add you as a contributor if that helps

bmacnaughton:
> I guess I'm confused then. To restate my preferences:
>
>Mocking ideally

sorry i wasn't more clear. i don't believe that it's a valid test to use a real proxy (test/proxy.util.js) combined with nock to mock requests to s3. it might work to set up nock to fulfill the requests coming out of the proxy but with nock patching the http code i'm not exactly sure what we're testing.

> If that does not work, writing to a real bucket, but not requiring that bucket to have state
I think its okay to have a test that only runs on contributor/not external PRs, as long as you can work with me to ensure it is passing.

this is certainly doable given that it won't run on external PRs.

> Happy to add you as a contributor if that helps

if you're comfortable doing that i'd be happy to get it working on my own.

springmeyer:
I've send you an invite as contributor - let me know if that works to get the tests working.

bmacnaughton:
i will get to this over the weekend. trying to get out an rc today.

springmeyer:
landed in #551

PR #551, opened 2021-01-30
Node fetch with proxy

bmacnaughton:
supersedes the PR bmacnaughton/node-fetch-with-proxy

new work
- replaces needle and request with node-fetch
- adds proxy support for both http and https destinations (but only http proxy)
- adds testing with builtin proxy server

user visible changes
- changes messages issued by install - the flow changed dramatically

additional changes
- proxy tests are stateless
- adds s3-bucket switching utility to allow testing against any developer's bucket

(i've got some work to do on getting the CI tests working)

seems like there is a race condition where multiple jobs across builds and CI platforms can interfere with one another. when one job running node 10 mocks s3 and another job running node 10 doesn't mock s3 then there is no conflict. but when the `travis ci - branch` and `travis ci - pull request` builds are running concurrently then can interfere with each other. and when the appveyor builds are not mocking s3 both the branch and pr builds increase the likelihood of hitting the race condition and causing a test to fail. the most common symptom i saw was a 403 (which is what s3 returns when a file isn't found) for one of actions being tested.

i came to this conclusion after the previously working travis jobs failed when i changed appveyor tests not to mock s3. i then set travis to run only one job at a time and the tests for that commit succeeded.seems like there is a race condition where multiple jobs across builds and CI platforms can interfere with one another. when one job running node 10 mocks s3 and another job running node 10 doesn't mock s3 then there is no conflict. but when the `travis ci - branch` and `travis ci - pull request` builds are running concurrently then can interfere with each other. and when the appveyor builds are not mocking s3 both the branch and pr builds increase the likelihood of hitting the race condition and causing a test to fail. the most common symptom i saw was a 403 (which is what s3 returns when a file isn't found) for one of actions being tested.

i came to this conclusion after the previously working travis jobs failed when i changed appveyor tests not to mock s3. i then set travis to run only one job at a time and the tests for that commit succeeded.

@springmeyer

- this can be minimized by limiting concurrent jobs to 1 though there is still a window where the race condition can be hit because there could be 4 builds (2 travis, 2 appveyor) executing concurrently.
- it's possible to change the production code back to having no awareness of mocking **if it's ok to let external PR CI builds fail**. i can use `test/switch-s3-hosts.js` to brute-force change code from referencing the mapbox bucket to reference a bucket that i have write access to and that allows all the tests to be run locally. it does require a contributor to have an s3 account and that they change `switch-s3-hosts.js` to reference their bucket instead of the one i created in my account.

more to come


bmacnaughton:
i was able to use AWS4 to sign requests relatively quickly but i have given up on getting it to work through a CONNECT proxy. The problem is that AWS4 signatures (and maybe previous versions of signatures, but i didn't check) and the CONNECT proxy-agent and proxy incompatible. It might be possible to modify the proxy-agent to do the AWS4 signatures but it's not obvious (to me at least) how that needs to work to meet AWS expectations.

I spent a fair amount of time on it, including modifying the http-proxy-agent code to sign the requests after it fiddled with the path and host; it's possible that i just missed an obvious solution because i was working with in new territory. but i don't have a clear path to implementation at this point and, not knowing exactly what AWS expects once a connection has been established, not sure how to get there. i did find one [2014 post](https://github.com/aws/aws-sdk-js/issues/108) where it noted the SDK does not support CONNECT proxies but they appear to do so now. maybe a future effort when i have time to look through the aws-sdk source code.

i'm going to work on getting this cleaned up but without trying to replace aws-sdk with pure http.

bmacnaughton:
the code coverage decrease appears to be due not using the mock-s3 code during testing. if moving the testing back to use a real s3 then the mock-s3 code can be removed.

the `test/switch-s3-hosts.js` utility is a bit of hack; it just goes through the code tree and changes references to
`https://mapbox-node-pre-gyp-public-testing-bucket.s3.us-east-1.amazonaws.com` to my bucket, `https://bmac-pre-gyp-test.s3.us-east-1.amazonaws.com` (and back).

it would be nice to allow the substitute s3 destination to be taken from the environment, but because of the need to quote elements of the regex expression for the `toMapbox` it seems more straightforward to just make it clear in the code/readme.

@springmeyer - what do you think of removing the mock-s3 code and adding a note in the contributing document that says something like "use test/switch-s3-hosts.js to enable testing your changes; see the file for details"?

editing the test files is a bit of a kludge but seems like a better solution that having the run-time mock-s3 code in place. if you're ok with it, i'll make those changes in this. (all tests complete without error locally when using switch-s3-hosts to change to a bucket i have access to.)

bmacnaughton:
if not removing the s3 mocking then i think this is ready to go. if removing s3-mocking then there should be a note in `contributing.md` that lets people know that the CI tests will fail because they don't have write access to the buckets.

bmacnaughton:
@springmeyer - just making sure you see this.

test/switch-s3-hosts.js
```diff
@@ -0,0 +1,76 @@
...
+// if you are not a collaborator to the mapbox/node-pre-gyp repository).
+//
+// this replaces the mapbox-specific s3 URLs with an URL pointing to an S3
+// bucket which can be written to. each person using this will need to supply
+// their own `toLocal.target` and `toMapbox.source` values that refer to their
+// s3 buckets (and set AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
+// appropriately).
+//
+// reset to the mapbox settings before committing.
+//
```
springmeyer, line 17: This looks like a utility script rather than a test, correct? If so, and I'm understanding right, how about moving this to the `scripts/` directory?

test/proxy.test.js
```diff
@@ -0,0 +1,415 @@
+'use strict';
+
+const test = require('tape');
```
springmeyer, line 3: This test file looks like it was copied from `test/build.test.js` and modified. I presume you did this because it is different enough that it does not make sense to integrate into `test/build.test.js`? If so, can you at least clean up the commented code in this file to avoid future confusion of whether that commented code is intended to work?

appveyor.yml
```diff
@@ -1,8 +1,8 @@
 os: Visual Studio 2015
 
 environment:
-  global:
-    node_pre_gyp_mock_s3: true
+  #global:
+  #  node_pre_gyp_mock_s3: true
```
springmeyer, line 5: Can this commented code be removed now since we declare the right env per job and the defaults are reasonable?

springmeyer:
@bmacnaughton I'm struggling to have an opinion myself about keeping mocking or not. My take is that the script to change buckets, while a hack, is useful so okay to keep. But I think my additional take is that keeping mocking feels useful to allow PRs to pass for external contributors. I made a few inline comments asking for code to be cleaned up (commented code to be removed to avoid confusion). After that can you ping me again for review? This feels close to being mergable and then part of next v 1.0.0 alpha tag. Thanks!

test/proxy.test.js
bmacnaughton: i am happy to remove the commented out code; the reason i left it there was to keep the code as parallel as possible for future comparison. if that doesn't seem worthwhile to you then i'll clean it up.

my theory was that testing the proxy with a single test app was good enough; if that is a bad assumption i can just uncomment the additional tests and work on merging this file into `build.test.js`.

appveyor.yml
bmacnaughton: i missed these when uncommenting the travis.yml definitions. i think i need to uncomment them so they won't fail on an external PR. i can test by creating an external PR that just uncomments them; if they don't fail then i'll remove them.

test/switch-s3-hosts.js
bmacnaughton: moved.

appveyor.yml
bmacnaughton: i guess i was missing something. removed.

test/proxy.test.js
bmacnaughton: @springmeyer - this is a bit of an open question for me. the other items are cleared.

springmeyer: 👍  seems fine to test with just one app. So, seems okay to just remove the commented code and simplify the test code as much as possible.

bmacnaughton: removed

bmacnaughton:
@springmeyer - changes made. let me know of anything else that you spot!

springmeyer (approved):

Issue #535, opened 2021-01-05
Outstanding Security Vulnerability via RC which is using an out of date INI version


martin-fogelman:
Hi Folks,

Node-pre-gyp uses RC, which in turn is using an out of date ini version with a high severity prototype pollution vulnerability: https://app.snyk.io/vuln/SNYK-JS-INI-1048974

It looks like RC hasn't been updated in some time and it's already been a few weeks (granted there were holidays), so I'm escalating the issue here in case folks can help. See [RC issue 120](https://github.com/dominictarr/rc/issues/120) and RC [resolved but unmerged PR 121](https://github.com/dominictarr/rc/pull/121).

A quick review does seem to show that Dominic was responsive in the past re minimist? (See [RC pull 114](https://github.com/dominictarr/rc/pull/114) and [RC pull 115](https://github.com/dominictarr/rc/pull/115); cf. #493)

Thanks

springmeyer:
I think node-pre-gyp can likely get away without using RC in a future release. So I may just go that direction: https://github.com/mapbox/node-pre-gyp/tree/remove-rc

jmz527:
@springmeyer just checking in, any updates on this?

springmeyer:
Yes, node-pre-gyp upcoming v1.0.0 release will drop rc (#552) so this issue will be resolved

motishani:
@springmeyer do you know when the new v1.0.0 will be released?

springmeyer:
@motishani an alpha is already available. Try doing:

```
npm install @mapbox/node-pre-gyp@1.0.0-alpha4 --save
```

springmeyer:
@mapbox/node-pre-gyp@1.0.0 is now released, which solves this.

PR #552, opened 2021-01-31
Drop rc dependency

springmeyer:
This dependency is barely used and it is easy to get by without it. Addresses #535

However this removes the s3 host env support just because it was using rc. Not sure if that is absolutely critical to keep or not. /cc @bmacnaughton


CHANGELOG.md
```diff
@@ -2,13 +2,13 @@
...
-- Added back support for symlinks inside the tarball
+- New: support for staging and production s3 targets (see README.md)
+- BREAKING: no longer supporting `node_pre_gyp_accessKeyId` & `node_pre_gyp_secretAccessKey`, use `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` instead to authenticate against s3 for `info`, `publish`, and `unpublish` commands.
 - Dropped node v6 support, added node v14 support
 - Switched tests to use mapbox-owned bucket for testing
 - Added coverage tracking and linting with eslint
+- Added back support for symlinks inside the tarball
 - Upgraded all test apps to N-API/node-addon-api
-- New: support for staging and production s3 targets (see README.md)
-- Added `node_pre_gyp_s3_host` env var which has priority over the `--s3_host` option or default.
```
bmacnaughton, line 11: i wasn't thinking get rid of the env var completely; it's still important to override the default without modifying the command line(s) in package.json.

is a new name preferred?

lib/node-pre-gyp.js
```diff
@@ -255,13 +254,7 @@ proto.setBinaryHostProperty = function(command) {
   if (command === 'publish') {
     target = 'staging_host';
   }
-  // each test is separate as opposed to combined like "if (this.rc[x] || this.opts[x])" to give
-  // priority to env var settings over command line.
-  if (this.rc['s3_host'] === 'staging') {
-    target = 'staging_host';
-  } else if (this.rc['s3_host'] === 'production') {
-    target = 'production_host';
-  } else if (this.opts['s3_host'] === 'staging') {
```
bmacnaughton, line 264: related to previous point; i'll plan on a commit that just checks the env var in this section.

bmacnaughton (approved):
looks good other than removing the note about the env var from the doc/changelog.

lib/node-pre-gyp.js
springmeyer: 👍  per email, please add as a followup. So I'll merge this as-is and let you fix up with `process.env` usage as you see fit (I don't have opinions on the naming)

PR #564, opened 2021-02-13
skip tests requiring s3 if external PR

bmacnaughton:
i think this lets external PRs succeed by skipping tests that require S3 access to mapbox's buckets.

i will request a review after testing as an external PR as well.


bmacnaughton:
i think this is good now.

springmeyer (approved):
looks great, thank you

Issue #348, opened 2018-03-02
node-pre-gyp doesn't handle arm architectures properly.


nicolasnoble:
The "arm" target in itself isn't enough. If you look at [node binary distributions](https://nodejs.org/dist/v8.9.4/), you have 3 variants for arm: `armv6l`, `armv7l`, and `arm64`.

An example of an `armv6l` platform is a raspberry pi zero. Installing nodejs on a raspberry pi zero through nvm will yield the correct download of the correct binary (armv6l).

An example of an `armv7l` platform is a raspberry pi 3. Installing nodejs on a raspberry pi 3 through nvm will also yield the correct download of the correct binary (armv7l).

However, for both platforms, node-pre-gyp identify them as "arm", meaning we can only cross compile and publish a single version of node-pre-gyp packages, either armv6l or armv7l, which aren't interchangeable due to dynamic library versions and tags.

This creates a problem where users will be able to download the wrong version of the package for their platform, and crash on load. The only method for them to fix it is to uninstall the node-pre-gyp enabled package, and re-install it using a flag to force falling back on ignoring published packages and compile from source instead, which is not obvious.

Also, the documentation states that the only correct values for --target_arch are ia32, x64 and arm, whereas arm64 is in fact a valid keyword to use here. The produced packages work properly on an arm64 environment.

nicolasnoble:
Note that testing these platforms requires only one raspberry pi zero, and one raspberry pi 3. The raspberry pi zero can then be used to have an armv6l architecture using the typical raspbian zero distribution, while the raspberry pi 3 can be used for both the armv7l architecture using the typical raspbian distribution, and the arm64 architecture using the [pi64 distribution](https://github.com/bamarni/pi64).

nicolasnoble:
After digging the code a bit, I realized that the --target_arch argument isn't actually mapped to anything special, and that passing arm64 there should work properly. After experimenting a bit, it appears it does indeed work properly when cross compiling for arm64.

springmeyer:
>  I realized that the --target_arch argument isn't actually mapped to anything special, and that passing arm64 there should work properly.

👍 correct. But it will default to `ia32, x64, etc` when these can be detected by using `process.platform` from node.js core.

nicolasnoble:
I'm sorry, but that issue wasn't to be closed. While arm64 kinda works, armv6l still doesn't.

springmeyer:
sorry, re-opening

nicolasnoble:
[...]

PR #566, opened 2021-02-15
fix for armv6 / raspberry pi zero

szh:
See https://github.com/mapbox/node-pre-gyp/issues/348#issuecomment-385292205

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -280,6 +280,12 @@ module.exports.evaluate = function(package_json, options, napi_build_version) {
   const v = package_json.version;
   const module_version = semver.parse(v);
   const runtime = options.runtime || get_process_runtime(process.versions);
+  
+  let arch = process.arch
+  if (arch === 'arm') {
+    arch = 'armv' + process.config.variables.arm_version + 'l'
+  }
+
   const opts = {
     name: package_json.name,
     configuration: options.debug ? 'Debug' : 'Release',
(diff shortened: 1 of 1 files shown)


springmeyer (changes requested):
@szh thanks for providing this. However there are two blocking things:

 - This will need tests
 - This is backwards incompatible: any existing installs (for binaries already published) will be broken and uninstallable if they upgrade to this. So, you'll need to think through a way to handle this without the risk of breaking existing installs

szh:
I don't have the time to put into this, so you can close this if you want - unless someone else wants to take a crack at it

springmeyer:
okay, will close then, thanks

duhruh:
@springmeyer trying to get caught up on this, as i'm running into issues with this too now. Trying to brainstorm solutions to this, is there any reason we couldn't specifically check for `armv6`  something like

```
 let arch = process.arch
  if (arch === 'arm' && process.config.variables.arm_version === 6) {
    arch = 'armv6l'
  }
```

or would this also break existing binaries? If that's the case then the only other thing I can think of right now would be to put this behind a feature flag something like `process.env.WITH_ARM_VERSION`  I'm willing to propose the PR if you think this is viable.

Thoughts?

springmeyer:
@duhruh I'm pretty sure that would/could break existing installs because that same code is used to resolve already published binaries - so if folks have binaries with just `arm` in them they'd all of a sudden start looking for `armv6l` for previous releases and those would not be found.

> If that's the case then the only other thing I can think of right now would be to put this behind a feature flag something like process.env.WITH_ARM_VERSION I'm willing to propose the PR if you think this is viable.

Yeah, something like that seems required (if a little clumsy). I wonder if the feature flag could be more generic to be more future proof: does this problem exist for other architectures such that the feature flag would not have to be arm specific but rather named something like `WITH_PROCESSOR_VERSIONING` or something?

PR #567, opened 2021-02-22
Respect npm_config_arch for cross-compilation scenarios

dennisameling:
Electron supports cross-compilation (e.g. from x64 to arm64) through the `npm_config_arch` environment variable: https://www.electronjs.org/docs/tutorial/using-native-node-modules#using-npm

`prebuild-install` checks the target architecture [in a three-step process](https://github.com/dennisameling/prebuild-install/blob/43d581a7225d1e922715b1d1e84ef98ffbcd1033/rc.js#L17):
1. `pkgConf.arch` (through running `prebuild install --arch XXX`
2. `env.npm_config_arch` for cross-compilation scenarios (e.g. a x64 host targeting arm64)
3. `process.arch` for the architecture that Node is currently running on

This PR also brings that behavior to `node-pre-gyp` 🚀

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -300,8 +300,8 @@ module.exports.evaluate = function(package_json, options, napi_build_version) {
     target: options.target || '',
     platform: options.target_platform || process.platform,
     target_platform: options.target_platform || process.platform,
-    arch: options.target_arch || process.arch,
-    target_arch: options.target_arch || process.arch,
+    arch: options.target_arch || process.env.npm_config_arch || process.arch,
+    target_arch: options.target_arch || process.env.npm_config_arch || process.arch,
     libc: options.target_libc || detect_libc.family || 'unknown',
     module_main: package_json.main,
     toolset: options.toolset || '' // address https://github.com/mapbox/node-pre-gyp/issues/119


springmeyer:
@dennisameling thanks. A few questions:

 - Is this a feature enhancement or a regression fix?
 - This will need a test to ensure it does not regress before merging. Is that something you can work on adding to this PR?

dennisameling:
- I guess a feature enhancement. I was just working on a repo which uses `node-pre-gyp` and noticed that it uses `ARM64` in the build process, but the generated binary ended in `x64`. So I guess it has never worked in scenarios that use `npm_config_arch`
- Sure, will add the test later today or tomorrow 👍🏼

springmeyer:
> Sure, will add the test later today or tomorrow

👍  thanks.

dennisameling:
Just a quick note that I didn't get to writing the test today - have rescheduled for tomorrow. Will keep you posted, sorry for the delay

dennisameling:
@springmeyer just added the test. The weird thing is if that I run `tape test/*test.js` the test passes, but if I run `npm run test` it doesn't. Looks like `stderr` is empty in the latter case. Do you have any idea why that could be? Thanks in advance!

springmeyer:
> @springmeyer just added the test. The weird thing is if that I run tape test/*test.js the test passes, but if I run npm run test it doesn't (am on Windows). Looks like stderr is empty in the latter case. Do you have any idea why that could be? Thanks in advance!

Bizarre, no. Are all your tests uniquely named?

> Also, Linux doesn't support cross-compilation unless things like crossbuild-essential-${ARCH} are installed. It does certainly work on Windows MSVC and I think on MacOS as well. Shall I only run the tests on Windows for now?

Right. Hmm. Yeah, would be nice to test this without compiling. Perhaps you could just have a test that runs `node-pre-gyp install --loglevel=http` and while that would fail you could assert on the output of path and that it contains the right target platform.

Or instead add a test that looks at the versioning specifically without messing with a build: https://github.com/mapbox/node-pre-gyp/blob/master/test/versioning.test.js

Issue #572, opened 2021-03-17
Unable to use proxy authentication


feuxfollets1013:
## Current Behavior 
I set auth-proxy address to npm proxy as following and I got invalid proxy error when I built some library.

- npm proxy setting

```
https-proxy=http://id:pass@example.com:18080/
proxy=http://id:pass@example.com:18080/
```

- error

`node-pre-gyp WARN download ignoring invalid "proxy" config setting: "http://id:pass@example.com:18080/"`

## Expected Behavior
The build process uses proxy config setting.

## Additional Info
In the following code(L63), it looks like auth-proxy and 5 digits port is not taken into account.
I checked whether node-https-proxy-agent(L65) supports auth-proxy, and I guess it supports auth-proxy from this [comment](https://github.com/TooTallNate/node-https-proxy-agent/issues/12#issuecomment-216098644).
https://github.com/mapbox/node-pre-gyp/blob/f9b39484f17955d83cdab42c178a600467fe96bd/lib/install.js#L63-L75

springmeyer:
That code is new in the v1.x version of node-pre-gyp. So I presume this worked for you with node-pre-gyp@v0.12.0? Asking to try to understand if this is truly a regression or not.

/cc @bmacnaughton

feuxfollets1013:
Thank you for your reply.
I suppose it is regression.  I tried building [bcrypt](https://github.com/kelektiv/node.bcrypt.js) @v.5.0.0 and @v.5.0.1.  
According to the change log, bcrypt has only changed node-pre-gyp@0.15.0 to @1.0.0, when bcrypt updated from v.5.0.0 to v5.0.1.

The building result is following.
- v.5.0.0: success
- v.5.0.1: fail `WARN download ignoring invalid "proxy"`

bmacnaughton:
i will take a look this weekend. @feuxfollets1013  - thank you for the report.

bmacnaughton:
@feuxfollets1013  - can you install `npm install https://github.com/mapbox/node-pre-gyp#proxy-url-work and see if that works for you?

i probably should always have let the agent parse the proxy url; i've changed it to do so.

feuxfollets1013:
@bmacnaughton - Thank you for your support.
I will run it next Monday.
As far as I can see, the code works because if proxyUrl is malformed, then the ProxyAgent constructer throws an error.

soulchild:
[...]

PR #573, opened 2021-03-19
Proxy url work

bmacnaughton:
wip - draft


springmeyer:
Given https://github.com/mapbox/node-pre-gyp/issues/572#issuecomment-808601263 happy to approve this and push it out anytime. But, would it be possible to add tests for any of the conditions found in #572? ( I realize maybe not since we can't easily test live proxies).

bmacnaughton:
let's give the two reporters a chance to weigh in (a few days).

on the test - i'm thinking of adding a bcrypt installation test because it uses https and has a redirect involved. it's really painful to set up a "real" test with https because you either need a real cert or must turn off node's cert validation. in case it's real money and effort and in the other case we're not really testing the normal node-path. i'll take a look at that.

bmacnaughton:
the single failure in the previous appveyor run is because github.com could not be resolved; regardless, i wanted a clean build and found something to change and generated a gratuitous commit. based on testing by @soulchild and @feuxfollets1013, i believe that this is ready for release.

test/proxy-bcrypt.test.js
```diff
@@ -0,0 +1,160 @@
...
+    // if no errors on download and the file is there that's good enough. napi version
+    // differences make it harder to select a version that is loadable, and this test
+    // is focused on proxy support.
+    .then(() => {
+      t.equal(expectedCount, 1, 'should find the expected file in the tar stream');
+      t.doesNotThrow(() => fs.statSync(withDir));
+      t.end();
+    })
+    .catch((e) => {
+      console.log(e);
```
springmeyer, line 144: Is this `console.log` left over from debugging and can be removed or is it needed for the tests?

test/proxy-bcrypt.test.js
```diff
@@ -0,0 +1,160 @@
...
+      t.end();
+    });
+  });
+
+
+});
+
+test('verify node fetch with a proxy successfully downloads bcrypt pre-built', (t) => {
+  // "{module_name}-v{version}-napi-v{napi_build_version}-{platform}-{arch}-{libc}.tar.gz"
+  const url = 'https://github.com/kelektiv/node.bcrypt.js/releases/download/v5.0.1/bcrypt_lib-v5.0.1-napi-v3-linux-x64-glibc.tar.gz';
```
springmeyer, line 95: /cc @kelektiv - so that you are aware that we're building logic into our tests that reference your downloads. Do you have any concern with this? (e.g. hopefully you don't plan to remove files like this in the future and rather they will stay around indefinitely?)

test/proxy-bcrypt.test.js
```diff
@@ -0,0 +1,160 @@
...
+const proxyPort = 8124;
+const proxyServer = `http://localhost:${proxyPort}`;
+// options for fetch
+const options = {};
+
+let initial_s3_host;
+let initial_mock_s3;
+
+// the temporary download directory and file
+const downloadDir = 'download';
```
springmeyer, line 28: How about using `os.tmpdir()` instead of a local directory (that would need to be put into .gitignore and .npmignore)?

springmeyer (commented):
Looking great, just a few questions I dropped before approving.

test/proxy-bcrypt.test.js
bmacnaughton: will do

test/proxy-bcrypt.test.js
bmacnaughton: it's a test and seems worthwhile to output the error if there is one. before adding this, it was just `t.error(e)` but that didn't provide enough information to see the problem. i'd prefer to leave it in so that an error here provides some useful clues.

test/proxy-bcrypt.test.js
bmacnaughton: worst cases:
1. we can update the tests if they remove that particular release
2. we can add logic to find a recent release and fetch that; i wanted to defer that so we could get a fix out but it could be follow-on work if need be

test/proxy-bcrypt.test.js
bmacnaughton: done

springmeyer (approved):

springmeyer:
Released as `@mapbox/node-pre-gyp@1.0.2`

PR #576, opened 2021-04-05
Allow alternative s3 urls and s3-compatible backends (e.g. minio). 

### LINKED ISSUE #571: unable to publish to s3 - Cannot read property 'slice' of undefined
I recently upgraded to the latest verision 1.x from 0.x. Since then unpublish command is crashing

```
node-pre-gyp ERR! UNCAUGHT EXCEPTION 
node-pre-gyp ERR! stack TypeError: Cannot read property 'slice' of undefined
node-pre-gyp ERR! stack     at Object.module.exports.detect (/home/runner/work/lexactivator-js/lexactivator-js/node_modules/@mapbox/node-pre-gyp/lib/util/s3_setup.js:21:29)
node-pre-gyp ERR! stack     at unpublish (/home/runner/work/lexactivator-js/lexactivator-js/node_modules/@mapbox/node-pre-gyp/lib/unpublish.js:18:12)
node-pre-gyp ERR! stack     at Object.self.commands.<computed> [as unpublish] (/home/runner/work/lexactivator-js/lexactivator-js/node_modules/@mapbox/node-pre-gyp/lib/node-pre-gyp.js:81:37)
node-pre-gyp ERR! stack     at run (/home/runner/work/lexactivator-js/lexactivator-js/node_modules/@mapbox/node-pre-gyp/lib/main.js:81:30)
node-pre-gyp ERR! stack     at processTicksAndRejections (internal/process/task_queues.js:79:11)
node-pre-gyp ERR! System Linux 4.15.0-1108-azure
node-pre-gyp ERR! command "/opt/hostedtoolcache/node/12.21.0/x64/bin/node" "/home/runner/work/lexactivator-js/lexactivator-js/node_modules/.bin/node-pre-gyp" "package" "unpublish" "publish" "info"
```

I debugged and in your code it crashes at:

```
if (!config.region) {
    const region = parts[1].slice(1).split('.')[0]; // this line causes the crash
    if (region === 'amazonaws') {
      config.region = 'us-east-1';
    } else {
      config.region = region;
    }
  }

```

pive:
Fixes #571 and allows more options for configuring s3.
Backward compatible and tests passed on Appveyor and TravisCI.


README.md
```diff
@@ -161,6 +161,12 @@ It is highly recommended that you use Amazon S3. The reasons are:
 
 Why then not require S3? Because while some applications using node-pre-gyp need to distribute binaries as large as 20-30 MB, others might have very small binaries and might wish to store them in a GitHub repo. This is not recommended, but if an author really wants to host in a non-S3 location then it should be possible.
 
+It is also possible to use an s3-compatible storage backend, such as minio, as well as non-standard s3 endpoints. In this case, you can specify the following properties in the `binary` section:
+  - `host`: should be the root url of the storage server.
+  - `bucket`
+  - `region`
+  - `s3ForcePathStyle` [optional, default: `false`, must be set to `true` for minio servers]
```
springmeyer, line 168: Could you try to make this comment more generic? For example, there might be users that need to set this to `true` but are not using or familiar with `minio`: how would they know to set it to `true` / can you describe more detail about why it needs to be set to `true`?

README.md
```diff
@@ -161,6 +161,12 @@ It is highly recommended that you use Amazon S3. The reasons are:
 
 Why then not require S3? Because while some applications using node-pre-gyp need to distribute binaries as large as 20-30 MB, others might have very small binaries and might wish to store them in a GitHub repo. This is not recommended, but if an author really wants to host in a non-S3 location then it should be possible.
 
+It is also possible to use an s3-compatible storage backend, such as minio, as well as non-standard s3 endpoints. In this case, you can specify the following properties in the `binary` section:
```
springmeyer, line 164: Can you add a little more detail around what kind of errors or situations a user might hit such that they would need to provide these options?

springmeyer (commented):
Looks great, thank you. Added a few inline questions asking for more doc details. But otherwise this looks good to merge and release in next release.

README.md
pive: How about:
Set s3ForcePathStyle to true if the endpoint url should not be prefixed with the bucket name. If false (default), the server endpoint would be contructed as `bucket_name.your_server.com`.

springmeyer: 👍

README.md
pive: What about:
If you are not using a standard s3 path like `bucket_name.s3(.-)region.amazonaws.com`, you might get an error on `publish` because node-pre-gyp extracts the region and bucket from the `host` url. For example, you may have an on-premises s3-compatible storage server, or may have configured a specific dns redirecting to an s3 endpoint. In these cases, you can explicitely set the `region` and `bucket` properties to tell node-pre-gyp to use these values instead of guessing from the `host` property. The following values will be used in the binary section:

springmeyer: 👍

springmeyer (approved):

springmeyer:
Tagged as @mapbox/node-pre-gyp@1.0.3

PR #579, opened 2021-04-15
Move aws-sdk from devDeps to deps

dmattia:
aws-sdk is imported here: https://github.com/mapbox/node-pre-gyp/blob/26fac61a03e74b577604db3d0ab696ba22443578/lib/util/s3_setup.js#L76

This file is included in the npm bundle, so the package should list aws-sdk as a dependency.

Right now, this breaks `esbuild` builds that have this library as a dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "bin": "./bin/node-pre-gyp",
   "main": "./lib/node-pre-gyp.js",
   "dependencies": {
+    "aws-sdk": "^2.840.0",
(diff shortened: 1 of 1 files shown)


springmeyer:
This is intended to be in devDeps since applications using node-pre-gyp to install binaries don't need the aws-sdk and it is therefore only needed for publishing. For publishing the docs say to install it yourself like: https://github.com/mapbox/node-pre-gyp#1-add-new-entries-to-your-packagejson

PR #590, opened 2021-06-27
Fix issue 589

ronilan:


bmacnaughton (approved):
this looks good. and thanks for both the doc changes and test.

i think this makes sense and should be in the next release.

@ronilan  - i don't do releases here and the mapbox person that does in on leave for a period of time, so this isn't likely to get released quickly. as the code works as documented and there is a clear workaround, i'm guessing that's ok for a month or two.

ronilan:
@bmacnaughton thanks. No urgent.

PR #593, opened 2021-07-28
Differentiate between aix and ibmi in package versioning

### LINKED ISSUE #592: Node.js (and by extension node-pre-gyp) doesn't differentiate between AIX and IBM i
IBM i is a business operating system that includes an AIX-like environment called PASE. Until recently, Python on IBM i identified the operating system as `aix`. Python is used when building the Node.js binary to determine the operating system, so the Node.js `process.platform` on when called on IBM i returns `aix`.

The problem arises when `node-pre-gyp` uses `process.platform` in the creation of published package names. This means binaries built on IBM i for IBM i have names that identify the platform as `aix` like so: `odbc-v2.4.0-aix-ppc64-napi-v6.tar.gz`. IBM i binaries won't run on AIX, and vice-versa, so anyone using AIX to download these packages will get binaries not built for their systems.

Extra logic should be added to `lib/util/versioning.js` to check the return value of `os.type()` when the platform is identified as AIX. `os.type()` on IBM i will return `'OS400'` (the historical name for the operating system), which will allow node-pre-gyp to differentiate between AIX and IBM i. If `op.type()` returns `'OS400'`, then the platform and target_platform should be changed to `ibmi`.

markdirish:
Fixes #592

Signed-off-by: Mark Irish <mirish@ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Options include:
  - `--fallback-to-build`: fallback to building from source if pre-built binary is not available
  - `--target=0.4.0`: Pass the target node or node-webkit version to compile against
  - `--target_arch=ia32`: Pass the target arch and override the host `arch`. Valid values are 'ia32','x64', or `arm`.
- - `--target_platform=win32`: Pass the target platform and override the host `platform`. Valid values are `linux`, `darwin`, `win32`, `sunos`, `freebsd`, `openbsd`, and `aix`.
+ - `--target_platform=win32`: Pass the target platform and override the host `platform`. Valid values are `linux`, `darwin`, `win32`, `sunos`, `freebsd`, `openbsd`, `aix`, and `ibmi`.
 
 Both `--build-from-source` and `--fallback-to-build` can be passed alone or they can provide values. You can pass `--fallback-to-build=false` to override the option as declared in package.json. In addition to being able to pass `--build-from-source` you can also pass `--build-from-source=myapp` where `myapp` is the name of your module.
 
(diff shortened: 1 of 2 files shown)


markdirish:
I have tested this on a N-API package I maintain, and on IBM i it both:

1. Looks for a binary with `ibm` as the platform in the version name instead of `aix`
2. Packages binaries in tarballs with `ibmi` as the platform in the version name instead of `aix`

markdirish:
Appear that the only failing portion of the CI build is the comment block below mine that was indented incorrectly, but that I left untouched.

springmeyer:
Thanks for providing this fix @markdirish. Because this seems like it would break existing installs (it would change resolution of existing binaries that might expect `aix`, right?) I think I'll need to include this in a new major version of @mapbox/node-pre-gyp. Sound good?

markdirish:
I think that's right. If you update now, AIX users will be unaffected because they were `aix` before and will be `aix` after. IBM i users will go from `aix` to `ibmi`, so a user running and older version of `node-pre-gyp` will try to pull `aix`, while a package developer may have updated and now package `ibmi` binaries. The reverse could also be true.

Then again, I'm not sure how many IBM i binaries there are out that that use `node-pre-gyp` other than our (IBM i OSS team) packages, and we could always just update our packages to require a minimum of whatever version this change is added into, and then ship releases from that dependency change with the `ibmi` instead of `aix` identifier.

springmeyer:
Thanks for the additional context @markdirish. I also just had a chance to read https://github.com/mapbox/node-pre-gyp/issues/592. I'm now understanding that this is likely to not break anyone out there, so is probably safe in any upcoming release.

But I have one remaining question: why not just work with the node core developers to land a patch that properly updates node such that `process.platform()` returned what you'd expect it to?

PR #599, opened 2021-08-08
Fix documentation for aws key enviroment variables

codeplea:
The documentation was using the old environment variable names for AWS in the travis ci example. I fixed it to show the proper environment variable names.


springmeyer:
Thanks for noticing this problem and providing a fix.

Issue #607, opened 2021-09-22
`s3ForcePathStyle` works for publishing but not installation


jared-duo:
I am using LocalStack for a local S3 for verification, integration tests etc., so I am using `localhost` and other domains not part of the normal production AWS forms.

I have followed the configuration instructions [here](https://github.com/mapbox/node-pre-gyp#s3forcepathstyle) and publishing works perfectly, but the `hosted_path` property in the `lib/utils/versioning.js` doesn't take this parameter into account (we can see in [the PR that introduces the `s3ForcePathStyle` parameter](https://github.com/mapbox/node-pre-gyp/pull/576/files?file-filters%5B%5D=.js#diff-731dda4952e8957a5d9387cd5ed656a10a7a6fd6e22b1ab3ef12502a9917e03dR307) that the `evaluate` method in `versioning.js` accepts the parameter, but never uses it.)

I believe the `opts.hosted_path` needs to be altered to account for the bucket name being part of the path so that installation may also occur from a non-standard AWS path.

## Example:
Given a 
## `package.json`
```json
{
  ...
    "binary": {
    "module_name": "my-module",
    "module_path": "<some_path>",
    "remote_path": "/npm/{module_name}/addon/",
    "package_name": "{module_name}.tar.gz",
    "host": "https://localhost.localstack.cloud:4566",
    "region": "us-east-1",
    "bucket": "my-packages",
    "s3ForcePathStyle": true
  }
 ...
}
```
and an S3 bucket available at `https://localhost.localstack.cloud:4566/my-packages` an invocation of `node-pre-gyp publish` will place the tarball correctly in `https://localhost.localstack.cloud:4566/my-packages/npm/my-module/addon/my-module.tar.gz`, but an invocation of `node-pre-gyp install` will try to download the tarball from `https://localhost.localstack.cloud:4566/npm/my-module/addon/my-module.tar.gz` - notably without the bucket in the path.

PR #608, opened 2021-09-22
Make `opts.hostedPath` (and installation) respect the `s3ForcePathStyle` param

jared-duo:
This is an example of a potential fix for issue #607.

I'm sure I may be doing something non-idiomatically as I do not code for NodeJS too often.

### DIFF
--- a/lib/util/versioning.js
+++ b/lib/util/versioning.js
@@ -329,7 +329,11 @@ module.exports.evaluate = function(package_json, options, napi_build_version) {
   const package_name = package_json.binary.package_name ? package_json.binary.package_name : default_package_name;
   opts.package_name = eval_template(package_name, opts);
   opts.staged_tarball = path.join('build/stage', opts.remote_path, opts.package_name);
-  opts.hosted_path = url.resolve(opts.host, opts.remote_path);
+  if (opts.s3ForcePathStyle) {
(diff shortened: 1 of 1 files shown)


lib/util/s3_setup.js
```diff
@@ -33,6 +32,10 @@ module.exports.detect = function(opts, config) {
       }
     }
   }
+  // If we are using s3ForcePathStyle the bucket is part of the http object path
+  // (but not the S3 key prefix path). If we aren't this isn't as much of a consideration.
+  const bucket_path = config.s3ForcePathStyle ? `/${config.bucket}/` : '/';
```
jared-duo, line 37: Moving the below the if/else structure _does_ skip an early `return` on line 21.  I _think_ that early return is acting in light of an error condition so not having `config.prefix` set for it is fine, but I'm not 100% sure.

eiygy (approved):

benmccann:
https://github.com/mapbox/node-pre-gyp/pull/650 fixed https://github.com/mapbox/node-pre-gyp/issues/607. @jared-duo can you test and see if this PR is still needed or should we close it?

ronilan:
> #650 fixed #607. @jared-duo can you test and see if this PR is still needed or should we close it?

@benmccann ou can close it.

PR #600, opened 2021-08-08
Upgrade `tar` to mitigate security vulnerabilities

thekumar:
Upgrade tar dependency to v.6.1.6 to mitigate the following security issues fixed in 6.1.1 and 6.1.2.

https://www.npmjs.com/advisories/1770
https://www.npmjs.com/advisories/1771

PR #609, opened 2021-09-28
bump tar version to solve sec vulnerabilities

drjauss:
Here we bump the tar dependency version since it is presenting a security vulnerability considered high.


drjauss:
This previous PR is also related [https://github.com/mapbox/node-pre-gyp/pull/601](https://github.com/mapbox/node-pre-gyp/pull/601)

TheVaan:
Duplicate to #600 that already bumps tar to 6.1.6.

Issue #623, opened 2021-12-10
#620 is a breaking change


bcoe:
Hello @springmeyer, first off thank you for your open source contributions (_we use this library on some of our client libraries at Google, and appreciate your work)._

----

I notice the recent update #620, which upgrades `npmlog` to address a regex vulnerability. Unfortunately, this is technically a breaking change, because it drops support for Node 10.

~I noticed that `npmlog` isn't used too heavily, would you be open to switching to an alternate logger, e.g., pino?~

_Edit: reading through the code-base, it seems like swapping out the progress behavior would perhaps be a pain in the neck._

springmeyer:
@bcoe - thanks for the heads up - that has not been apparent to me. My intention has been to keep node v10 support for the time being (and only drop at a major release).

I've also considered dropping `npmlog` - it's heavier that likely what is needed. It is used because this project was initially modeled after node-gyp which used `npmlog`. I'd accept a PR changing the logger, but it would likely (as you say) be a bit painful/tedious due to how many tests depend on the output (and the tests are not particularly well written - what I can say the project is not the most modern).

In the meantime are you looking for a new 1.0.x release that restores node v10 support (somehow)?

bcoe:
> In the meantime are you looking for a new 1.0.x release that restores node v10 support (somehow)?

If we could figure out a way, this would be amazing. But it might be challenging because I think the latest `ansi-regex` forces `v12`.

I also reached to npm, one option would be perhaps floating a patch for `npmlog`, if npm is open to it (CC: @wraithgar).

---

Just an FYI, I moved my CI/CD to using this approach a few months ago:

```
- run: npm install --production --engine-strict 
```

It's been a great way to catch when a dependency drops a major node version, before it actually breaks behavior.

wraithgar:
I don't know if a patch is possible.  If there were semver-compatible updates for `ansi-regex` that were not vulnerable they would be installed already because npmlog uses `^` as its prefix.

[Wide-align moved its support to node12 also when fixing this](https://github.com/iarna/wide-align/commit/cab66e9369f9a71a8762f726f9ee3970f0e72384#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R26), which is one of the reasons this can't be fixed in v5 of npmlog.

Long term, moving off of npmlog is probably the best option, though I know that's a tall order and there is a limited amount of time in the universe.

bcoe:
[...]

PR #624, opened 2021-12-13
Downgrade npmlog + upgrade gauge to address #623

springmeyer:
This should address #623 

 - Still keeps at least ansi-regex@5.0.1 thanks to gauge@3.0.2 release
 - And therefore can maintain node v10 support

TODO:

 - Why were, before this PR, tests not failing on node v10 (if support regressed per #623)?


wraithgar:
> Why were, before this PR, tests not failing on node v10

I can answer this and save you some time: because this was a pre-emptive drop in support.  Nothing in the code was changed that would have broke in node10.

The only thing that would catch something like this would be the `--engine-strict` flag.

springmeyer:
Hmm, @wraithgar - I just tried, locally, running

`npm install --production --engine-strict`

against the `master` branch (without this PRs fix) and it works without errors:

```
added 54 packages from 21 contributors and audited 394 packages in 1.684s

3 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

$ npm ls gauge                            
@mapbox/node-pre-gyp@1.0.7 /Users/danespringmeyer/projects/node-pre-gyp
└─┬ npmlog@6.0.0
  └── gauge@4.0.0 
```

Any idea why that is not erroring? I would assume I'd get a non-zero return code from `npm install` or `npm ci` with `--engine-strict`

springmeyer:
Perhaps I need to add `engines` to the `package.json` of `node-pre-gyp`?

wraithgar:
It looks like `npm@6` is not checking engines when installing from a package-lock.  `npm@7` always checks.

bcoe:
:tada: thanks for taking on this work. I know I'm like the only person who runs tests with `--engine-strict` :laughing:, and that I can be a squeaky wheel.

springmeyer:
@bcoe - no problem, I'd like to be like you. I'm still struggling however to replicate any kind of error or failure with node v10 (or 8 for that matter) and the `master` code before this fix. Even with `npm@7`. Would you be up for providing a PR to this repo's `.travis.yml` that demonstrates how to detect with `--engine-strict` the problem you were able to catch with CI? So, a PR that fails and then would pass once this PR is merged in.

bcoe:
@springmeyer this is the approach I've been taking that catches things:

https://github.com/googleapis/nodejs-logging/blob/main/.github/workflows/ci.yaml#L22

bcoe:
@springmeyer I did some local testing too, and a couple things I noticed:

* you need to be running on the version of Node.js that's minmally supported (the engines strict error only throws for me if I'm running `npm i` on Node 10).
* I was initially having problems with the package lock on npm@6 (so, as @wraithgar suggests, `npm i npm@7 -g` might be smart).

Once I did both these things, I'm getting the following with a local check out of `node-pre-gyp`:

```bash
node-pre-gyp$ npm i -production --engine-strict
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for npmlog@6.0.0: wanted: {"node":"^12.13.0 || ^14.15.0 || >=16"} (current: {"node":"10.24.1","npm":"6.14.15"})
npm ERR! notsup Not compatible with your version of node/npm: npmlog@6.0.0
npm ERR! notsup Not compatible with your version of node/npm: npmlog@6.0.0
npm ERR! notsup Required: {"node":"^12.13.0 || ^14.15.0 || >=16"}
npm ERR! notsup Actual:   {"npm":"6.14.15","node":"10.24.1"}
```

bcoe (approved):
thanks for doing this work 😄

springmeyer:
Got it, thank you @bcoe - npm@7 + using node v10 allowed me to replicate (and it caught another incompatible dependency - action-walk).

springmeyer:
This is now published as `@mapbox/node-pre-gyp@1.0.8`

minherz:
Thanks for doing this work 👏 it allows us to avoid dropping Node 10 on `@google-cloud/profiler`

PR #625, opened 2021-12-13
fix(deps): downgrade to npmlog with support for Node 10

### LINKED ISSUE #623: #620 is a breaking change

bcoe:
A patch release of npmlog was released that addresses security issues, but supports Node 10.

Fixes #623


bcoe:
duplicate.

PR #627, opened 2021-12-17
 chore: update cnpm mirror

ChelesteWang:


benmccann (approved):
This is not something I have experience with, but as far as I can tell is a change that others have made: https://github.com/cnpm/cnpm/issues/361

PR #631, opened 2022-01-19
Upgrade to version 2 of detect-libc

lovell:
Hi, the latest [detect-libc](https://www.npmjs.com/package/detect-libc) has been modernised to use the non-blocking Node.js Report API where available (Node.js >= 12), falling back to use an efficient single child process with older versions.

There's an extensive integration test suite, e.g. https://github.com/lovell/detect-libc/actions/runs/1718463728

Should anyone be interested in the background to this, please see lovell/detect-libc#14


axrj (approved):
lgtm, thanks

Issue #633, opened 2022-01-24
New vulnerability 


opravil-jan:
Hi, 
in package node-fetch is vulnerability. Can you fix it please?
```

node-fetch  <2.6.7
Severity: high
node-fetch is vulnerable to Exposure of Sensitive Information to an Unauthorized Actor - https://github.com/advisories/GHSA-r683-j2x4-v87g

```
Thanks

cjwilsontech:
Created a PR to fix. https://github.com/mapbox/node-pre-gyp/pull/634

axrj:
Hi, this is fixed in v1.0.9.

Issue #614, opened 2021-10-20
Repair node-webkit testing


springmeyer:
Node webkit testing is currently broken on CI:

```
$ ./scripts/test-node-webkit.sh
--2021-10-20 22:33:29--  https://dl.nwjs.io/v0.50.2/nwjs-v0.50.2-linux-x64.tar.gz
Resolving dl.nwjs.io (dl.nwjs.io)... 34.206.150.92
Connecting to dl.nwjs.io (dl.nwjs.io)|34.206.150.92|:443... connected.
ERROR: cannot verify dl.nwjs.io's certificate, issued by ‘CN=R3,O=Let's Encrypt,C=US’:
  Issued certificate has expired.
To connect to dl.nwjs.io insecurely, use `--no-check-certificate'.
The command "./scripts/test-node-webkit.sh" exited with 5.
```

I don't have the bandwidth to dig into this, so I'm going to disable it.

PR #637, opened 2022-03-10
use mock s3 for tests

axrj:
- Fixes currently failing tests by using s3 mock since access to S3 bucket is restricted for security reasons
- Adds variable to skip s3 tests for node 16. This should be investigated and fixed later with #638 
- Upgrade few dependencies especially node-fetch which fixes a security vuln #633 
- Removed proxy.test.js since the file is a near duplicate of build.test.js but using a proxy to access s3 for external forks which is now not required since tests use mocked s3
- Adds node-webkit test job back as it was disabled in #614


.travis.yml
```diff
@@ -30,38 +26,41 @@ matrix:
...
       node_js: 14
-    # Test node v16
+      env:
+        - node_pre_gyp_mock_s3=true
     - os: linux
       node_js: 16
-    # Test node v8
+      env:
+        - node_pre_gyp_mock_s3=true
+        - bypass_s3_tests=true
```
axrj, line 45: This is required for now so the tests which weren't running before won't fail and block travis ci. will create a new issue to tackle this.

.travis.yml
```diff
@@ -30,38 +26,41 @@ matrix:
...
-    #   node_js: 10
-    #   addons:
-    #     apt:
-    #       sources: [ 'ubuntu-toolchain-r-test' ]
-    #       packages: ['xvfb','libasound2','libx11-6','libglib2.0-0','libgtk2.0-0','libatk1.0-0','libgdk-pixbuf2.0-0','libcairo2','libfreetype6','libfontconfig1','libxcomposite1','libasound2','libxdamage1','libxext6','libxfixes3','libnss3','libnspr4','libgconf-2-4','libexpat1','libdbus-1-3','libudev1']
-    #   script:
-    #     # test node-webkit usage
-    #     - ./scripts/test-node-webkit.sh
+    - os: linux
+      node_js: 10.23 # check concurrency warning message at the top
```
axrj, line 56: use this specific version to have concurrency in travis jobs, otherwise it takes too long for all of them to finish

PR #648, opened 2022-05-04
Fix 638  Test Failing with Node 16

### LINKED ISSUE #638: S3 tests fail with node 16
The [tests which use S3](https://github.com/mapbox/node-pre-gyp/blob/master/test/build.test.js#L280) to publish and unpublish binaries weren't running on travis for the past few months. They can be activated currently when mock s3 flag is toggled.
"app can be installed via remote" test fails with node-gyp rebuild errors on node16 but passes with all older versions. The failure is not necessarily related to mock s3 library as they also fail with the same error when a dummy aws access key is provided in the variable.
This needs investigation and a fix.

Env: observed on travis jobs and local ubuntu bionic container
Node version : 16

ronilan:
### Overview

This pull request fixes #638.

### Change

- Removed `clang` flag from run command. Was breaking tests when run with npm 7.x and above (node 16, 18). See [issue comment](https://github.com/mapbox/node-pre-gyp/issues/638#issuecomment-1116815545) for error details.
- Removed `env.bypass_s3_tests` conditional from tests. No longer needed.
- Updated travis workflow no longer bypassing s3 tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,6 @@ matrix:
       node_js: 16
       env:
         - node_pre_gyp_mock_s3=true
-        - bypass_s3_tests=true
       addons:
         apt:
           sources: ['ubuntu-toolchain-r-test']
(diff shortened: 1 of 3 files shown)


axrj (approved):
Thanks!

Issue #654, opened 2022-05-17
Can't use Alternative S3 URLs with Staging/Production Option


ronilan:
## Overview

It is not possible to use alternative s3 urls (path style) and s3-compatible backends (e.g. min.io) in conjunction with the `production_host` and `staging_host` option.

## Status

- Originally `host` was defined as a string property of `binary`. The required AWS settings were automatically detected and extracted from the string.
- Pull request #533 defined `production_host` and `staging_host` as string properties of `binary` and the logic that governs which is used when. The required AWS settings were also automatically detected and extracted from the strings.
- Pull request #576 added `bucket` `region` and `s3ForcePathStyle` as string properties of the `binary` object. The trio of keys, along with `host` value, enables to explicitly set access to a bucket, bypassing automatic detection and extraction. 

## Problem 
`bucket` `region` and `s3ForcePathStyle` are set as properties of `binary` when in fact they are (together with the endpoint) properties of the host target. Since there can be multiple host targets (staging/production) but there can only be a single explicitly set bucket name, the alternate host option doesn't work in conjunction with the `production_host` and `staging_host` option.

## Proposed Solution

Moving forward host will be defined as an object with optional properties.

#### Example 1: (S3 Virtual Host URL)
```
{
  "binary": {
    "host": {
      "endpoint": "https://npg-mock-bucket.s3.us-east-1.amazonaws.com",
    }
  }
}
```
#### Example 2: (Path Style alternative host)
```
{
  "binary": {
    "host": {
      "endpoint": "https://play.min.io",
      "bucket": "node-pre-gyp-production",
      "region": "us-east-1",
      "s3ForcePathStyle": true
    }
  }
}
```
#### Example 3: (Utilizing Staging/Production)
```
{
  "binary": {
    "staging_host": {
      "endpoint": "https://npg-mock-bucket.s3.us-east-1.amazonaws.com",
    },
    "production_host": {
      "endpoint": "https://play.min.io",
      "bucket": "node-pre-gyp-production",
      "region": "us-east-1",
      "s3ForcePathStyle": true
    }
  }
}
```
```
{
  "binary": {
    "staging_host": {
      "endpoint": "https://play.min.io",
      "bucket": "node-pre-gyp-staging",
      "region": "us-east-1",
      "s3ForcePathStyle": true
    },
    "production_host": {
      "endpoint": "https://play.min.io",
      "bucket": "node-pre-gyp-production",
      "region": "us-east-1",
[...]

Issue #653, opened 2022-05-13
Chaining publish and unpublish Commands Changes Their Default Behavior


ronilan:
## Describe the bug

Chaining `publish` and `unpublish` commands changes their default behavior in relation to staging/production hosts.

## Expected Behavior
Commands `publish` and `unpublish` should **always** default to staging when `production_host` and `staging_host` are defined under `binary` in package.json.

## Actual Behavior

When the commands `publish` and `unpublish` are executed by themselves they default to staging. However, if the command being executed is chained after another command (for example `package publish`  or `info unpublish`) then the default reverts to `binary.production_host`. As a result accidental publishing/unpublishing is possible.

### 1) Prep package.json
Replace "host" with `staging_host` and `production_host` in package.json

Example:
```
"staging_host": "https://apm-appoptics-bindings-node-dev-staging.s3.us-east-1.amazonaws.com",
"production_host": "https://apm-appoptics-bindings-node-dev-production.s3.us-east-1.amazonaws.com",
```

### 2) Run Commands one after the other
`node-pre-gyp package`
`node-pre-gyp publish`

Results in: ✅ 
```
[node-pre-gyp-test-app1] published to https://apm-appoptics-bindings-node-dev-staging.s3.us-east-1.amazonaws.com/node-pre-gyp/node-pre-gyp-test-app1/v0.1.0/Release/
```

`node-pre-gyp info`
`node-pre-gyp unpublish`

Results in:  ✅ 
```
[node-pre-gyp-test-app1] Success: removed https://apm-appoptics-bindings-node-dev-staging.s3.amazonaws.com/node-pre-gyp/node-pre-gyp-test-app1/v0.1.0/Release/node-v83-darwin-x64.tar.gz
```

### 3) Run Chained Commands
`node-pre-gyp package publish`

Results in: ❌ 
```
[node-pre-gyp-test-app1] published to https://apm-appoptics-bindings-node-dev-production.s3.us-east-1.amazonaws.com/node-pre-gyp/node-pre-gyp-test-app1/v0.1.0/Release/
```

`node-pre-gyp info unpublish`

Results in: ❌ 
```
[node-pre-gyp-test-app1] Success: removed https://apm-appoptics-bindings-node-dev-production.s3.amazonaws.com/node-pre-gyp/node-pre-gyp-test-app1/v0.1.0/Release/node-v83-darwin-x64.tar.gz
```
## Root Cause
[...]

PR #652, opened 2022-05-12
Create GitHub Actions Test Setup

ronilan:

PR #651, opened 2022-05-10
Refactor Mock and Test Setup

ronilan:
### Overview

This pull request refactors the S3 mocking setup (see: #540) simplifying implementation, separates the S3 tests from the build tests, introduces test apps using alternative bucket specification (see: #576) and production/staging hosts (see: #533) and enables easy switch between mock bucket and real bucket (given status of: #613).

This pull request comes "on top" of (i.e. includes changes from) #648, #649, #650.

#### Code
* Refactored mock setup (e3c9772)
    - mock modules moved to `lib/mock`.
    - mock modules are self contained and have a signature similar to that of other modules in the package.
    - simplified http mock to remove no longer needed interface.
    - modified http mock to support path style bucket access.
    - set http mock to work with default `npg-mock-bucket` bucket.
* Refactored test setup (9085b88)
    - separated s3 tests from build tests.
    - refactored fetch test and proxy-bcrypt test to work with refactored mock setup.
* Refactored test apps and increased test coverage (4572c86)
    - set hosts of test apps to point to mock bucket.
    - added app1.1 - identical to app1 but using production and staging binary host option.
    - added app1.2 - identical to app1 but using explicit host, region, bucket options.
* Added service script to switch test apps bucket from mock to S3 and vis versa. Removed previous one (c2c54c5)

#### Testing
- "Out of the Box" (repo clone) tests will skip interaction with S3. 
- Setting the environment variable `node_pre_gyp_mock_s3` to true will run them against the mock ([see: repo CI](https://app.travis-ci.com/github/mapbox/node-pre-gyp/jobs/569765077)).
- To run against a real bucket:
    - unset environment variable `node_pre_gyp_mock_s3`.
    - set environment variables `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`.
    - run `npm run bucket {name-of-your-us-east-1-bucket}`
- To switch back to mock:
    - set environment variable `node_pre_gyp_mock_s3` to true.
    - run `npm run bucket`
- Manually configuring regions other than `us-east-1` also works but switch is more work.
- Results of running `npm run coverage` differ depending on what bucket is used (real s3 or mock).

### Tests
- All tests pass in both real bucket and mock bucket setups (including node 18).

PR #655, opened 2022-05-19
Refactor Staging/Production Alternate Host Implementation

ronilan:
## Overview

This pull request refactors the implementation of production/staging hosts (see: #533). It includes minor feature changes and is backwards compatible with any previously valid configurations.

This pull request comes "on top" of (i.e. includes changes from) #652 (which in turn comes "on top" of #651, which is "on top" of #648, #649, #650). Work is in preparation for resolving issue #654.

#### Features
- Added `development_host` option. Becomes default option for `publish` `unpublish` when present.
- Changed behavior when alternate hosts are defined. Now `production_host` acts as alias to `host`. Defining `staging_host` or `development_host` is enough to default `publish` and `unpublish` away from production.
- An invalid `s3_host` option does not result in error and is instead silently ignored.
#### Bug Fix (#653)
- When a chain of commands that includes `publish` or `unpublish` and host is not specifically set via command line or environment variable, **all** commands in the chain default away from production.
#### Code
- Moved logic regarding host selection of host to versioning. This is where all user defined values from package.json are transformed into command options.
#### Tests
- Tests of feature where moved from `run.test.js` to `versioning.test.js`.
- Increased test coverage for various options of using the hosts.
- All tests pass.


cclauss:
Please rebase now that #900 has been merged and our GitHub Actions are passing again.
* #900

PR #663, opened 2022-07-23
Add arm64 to target_arch in readme

maxsimych:


maxsimych:
Small fix to make readme more clear

springmeyer:
@maxsimych thanks for improving the docs. It strikes me that the ideal approach would be to say that any value is valid that node supports. The list/naming is at https://nodejs.org/api/os.html#osarch. What do you think about amending this PR to simply link to that page?

maxsimych:
@springmeyer done

springmeyer (approved):

Issue #697, opened 2023-09-25
Security issue in node-fetch < 2.7.0


Vhndaree:
https://security.snyk.io/package/npm/node-fetch

PR #698, opened 2023-09-25
node-fetch version upgrade

Vhndaree:
Fixes https://github.com/mapbox/node-pre-gyp/issues/697

- According to https://security.snyk.io/package/npm/node-fetch, there are some security vulnerabilities in the package and the vulnerability is fixed on >v2.7.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "detect-libc": "^2.0.0",
         "https-proxy-agent": "^5.0.0",
         "make-dir": "^3.1.0",
-        "node-fetch": "^2.6.7",
+        "node-fetch": "^2.7.0",
         "nopt": "^5.0.0",
         "npmlog": "^5.0.1",
         "rimraf": "^3.0.2",
@@ -2905,9 +2905,9 @@
       "dev": true
     },
     "node_modules/node-fetch": {
-      "version": "2.6.7",
-      "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.6.7.tgz",
-      "integrity": "sha512-ZjMPFEfVx5j+y2yF35Kzx5sF7kDzxuDj6ziH4FFbOp87zKDZNx8yExJIb05OGF4Nlt9IHFIMBkRl41VdvcNdbQ==",
+      "version": "2.7.0",
+      "resolved": "https://registry.npmjs.org/node-fetch/-/node-fetch-2.7.0.tgz",
+      "integrity": "sha512-c4FRfUm/dbcWZ7U+1Wq0AwCyFL+3nt2bEw05wfxSz+DWpWsitgmSgYmy2dQdWyKC1694ELPqMs/YzUSNozLt8A==",
       "dependencies": {
         "whatwg-url": "^5.0.0"
       },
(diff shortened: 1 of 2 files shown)


marck283 (changes requested):
NPM's node-fetch package versions up to 3.2.10 are subject to both a _MaxListenersExceededWarning: Possible EventEmitter memory leak detected_ warning and a _Regular Expression Denial of Service_ bug, so I think that the node-fetch dependency should be upgraded to node-fetch@3.2.10 instead of node-fetch@2.7.0.

Vhndaree:
Hi @marck283 Thanks, that is a great idea but 
`node-fetch` changed to ES modules by default from v3 so if we want to stick with the similar pattern we can't install node-fetch > v2.7 as in the [commit](https://github.com/mapbox/node-pre-gyp/pull/698/commits/4f5cc4d7d9f507c43cc1542f69e99a1eaf78d710). However, I also find the possibility to use same export but with new >v2.7 we need to do a small change as in the [commit](https://github.com/mapbox/node-pre-gyp/pull/698/commits/e2f820360a6edf7e08e1a32b724c554e841d67ef).

I'm not sure if it is possible at the moment but testing some thing in the second commit.

marck283:
I understand your point but consider that solving the _MaxListenersExceededWarning_ warning will lead to less memory usage. As shown in these test runs (taken from [here](https://github.com/node-fetch/node-fetch/issues/1295#issuecomment-933796394)):

- without the patch:
```
memoryUsage.rss: 35.97 Mb ⇒ 96.75 Mb (Diff: 168.99%)
memoryUsage.heapTotal: 7.11 Mb ⇒ 49.87 Mb (Diff: 601.10%)
memoryUsage.heapUsed: 5.23 Mb ⇒ 29.92 Mb (Diff: 471.66%)
memoryUsage.external: 1.04 Mb ⇒ 16.76 Mb (Diff: 1512.99%)
memoryUsage.arrayBuffers: 0.03 Mb ⇒ 15.68 Mb (Diff: 50570.52%)
```

- with the patch:
```
memoryUsage.rss: 35.87 Mb ⇒ 47.92 Mb (Diff: 33.61%)
memoryUsage.heapTotal: 7.11 Mb ⇒ 7.62 Mb (Diff: 7.14%)
memoryUsage.heapUsed: 5.22 Mb ⇒ 6.86 Mb (Diff: 31.51%)
memoryUsage.external: 1.04 Mb ⇒ 8.10 Mb (Diff: 679.89%)
memoryUsage.arrayBuffers: 0.03 Mb ⇒ 0.21 Mb (Diff: 592.11%)
```
the memory usage cuts to half or more in all the tested cases.

When it comes to the _require()_ pattern, Node.js users that run on version 13.2 and above can import the module by using the dynamic import instruction. This is why I think you should change the _require()_ instructions to dynamic imports, so users that run on versions of Node.js that do not support ES6 modules should always be able to import this module using the standard _require()_. Furthermore, by using the dynamic import instructions you can upgrade the node-fetch package this module runs on to its latest version. To allow the usage of the dynamic imports, however, you will have to change something in the files containing the TypeScript compilation settings.
For example, the file tsconfig.json should be modified as this (changing the file extension back to JSON): [tsconfig.txt](https://github.com/mapbox/node-pre-gyp/files/12772913/tsconfig.txt)

marck283 (approved):

JasonMan34:
[I created a PR that bumps minor versions across all dependencies including node-fetch](https://github.com/mapbox/node-pre-gyp/pull/702)

PR #712, opened 2024-04-11
Update for node Node 22, update AppVeyor and tests

acalcutt:
This is a merge of #649 #711 and #709 as requested by https://github.com/mapbox/node-pre-gyp/pull/649#issuecomment-2049490190
* #649
* #711
* #709

This PR updates node-pre-gyp to support up to Node 21. 
1.) I updated abi_crosswalk.json to include node 21 with 'npm run update-crosswalk'  
2.) It adds test coverage for cases of unsupported/unknown targets from #649
3.) It adds building with msvs 2022 and get AppVeyor to pass, from #709

Known Issues
Some tests are failing in node 20 and 21 for what look like minor line ending issues or slight path differences (which seem like differences in the responses from the node versions used more than a real issue with node-pre-gyp)

**Testing**
I have been using an updated abi_crosswalk.json  in [@acalcutt/node-pre-gyp](https://www.npmjs.com/package/@acalcutt/node-pre-gyp) and have used it with [maplibre-native](https://github.com/maplibre/maplibre-native)  package up to node 20 (not to 21 only because we are only supporting lts releases)

**Note from my own fork testing**
Do not publish this from windows, it messes up the line endings in abi_crosswalk.json and makes the package fail on linux.

PR #713, opened 2024-04-15
Upgrade `tar` to address security vulnerability

iAmmar7:
Upgrade the `tar` dependency to the latest v7.0.1 to address the following security issue:

https://github.com/advisories/GHSA-f5x3-32g6-xq36

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,9 @@ os: Visual Studio 2015
 
 environment:
   matrix:
-    - nodejs_version: 10
-    - nodejs_version: 12
-    - nodejs_version: 14
+    - nodejs_version: 18
+    - nodejs_version: 20
+    - nodejs_version: 21
 
 platform:
   - x64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,6 +24,7 @@ install:
   - IF /I "%PLATFORM%" == "x64" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
   - IF /I "%PLATFORM%" == "x86" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
   - npm ci
+  - npm run update-crosswalk
   - npm test
 
 build: off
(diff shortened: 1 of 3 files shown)


cclauss:
Please add the line `- npm run update-crosswalk` after line 26 of the file `appveyor.yml` so we can see if your tests pass?

appveyor.yml
```diff
@@ -24,6 +24,7 @@ install:
   - IF /I "%PLATFORM%" == "x64" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
   - IF /I "%PLATFORM%" == "x86" CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
   - npm ci
+  - npm run update-crosswalk
```
cclauss, line 27: Great!  Now set lines 5 thru 7 of `appveyor.yml` to
```python
    - nodejs_version: 18
    - nodejs_version: 20
    - nodejs_version: 21
```
to match https://nodejs.org/en/about/previous-releases#release-schedule and your tests should be green.  ✅

iAmmar7: The tests would be green because of the newer Node versions, but are you sure I should remove the older versions from the tests?

iAmmar7: Well, it did not pass for Node 20 and 21.

bensquire:
I'm out of my depth here, but just to say we're seeing pressure to also update our packages because of the above tar dependancy.

bensquire:
Thanks @cclauss. Will look at what I need to tear out in turn :)

appveyor.yml
```diff
@@ -2,9 +2,9 @@ os: Visual Studio 2015
 
 environment:
   matrix:
-    - nodejs_version: 10
-    - nodejs_version: 12
-    - nodejs_version: 14
+    - nodejs_version: 18
+    - nodejs_version: 20
+    - nodejs_version: 21
```
benmccann, line 7: node 22 is out now
```suggestion
    - nodejs_version: 22
```

appveyor.yml
benmccann: looks like there's an explanation for that here: https://github.com/mapbox/node-pre-gyp/pull/712/files#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dcR11

cclauss:
Git conflicts -- Please rebase.

PR #707, opened 2023-12-15
Bump rimraf to 5.0.5 to fix DoS

pnappa:
Older versions of rimraf transitively depended on a package called inflight, which is no longer maintained, and current has a medium severity security vulnerability associated with it. https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116

Newer versions of rimraf rely on a newer version of glob, which no longer imports inflight.

Due to the differences in major version, rimraf has since changed their API. They now return promises (as opposed to callbacks), and as a result, we need to provide two functions to the .then() continuation to invoke the callback correctly (with the parameters in the correct order).

Issue #692, opened 2023-07-16
replace `rimraf` with native rm


jimmywarting:
NodeJS v14 have `rm` built in with `recursive` option. i think you should use that rather than depending on [14 other sub dependencies](https://npmgraph.js.org/?q=rimraf@3.0.2)

PR #709, opened 2024-04-11
run.util.js: Test building with msvs 2022 and get AppVeyor to pass

cclauss:
# To support newer versions of Node.js > v17
```
npm ci
npm run update-crosswalk
```
---
`test/run.util.js`
```diff
  if (process.platform === 'win32') {
-    final_cmd += ' --msvs_version=2015 ';
+    final_cmd += ' --msvs_version=2022 ';
  }
```
---
`appveyor.yml`
```diff
+ matrix:  # TODO: Fix the `configures with unparsed options` tests and remove this allow_failures matrix.
+   allow_failures:
+     - nodejs_version: 20
+     - nodejs_version: 21
```
* mapbox/node-pre-gyp#710

### AppVeyor test results: https://ci.appveyor.com/project/Mapbox/node-pre-gyp/history

@yhahn, @mapsam, @axrj, @rafaykh90, @ewanharris, @ronilan, @acalcutt Your reviews, please.

Issue #705, opened 2023-12-08
Vulnerability in inflight


izumo27:
This package depends on `inflight` and it is vulnerable.
https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116
```
$ npm ls inflight --omit=dev
@mapbox/node-pre-gyp@1.0.11 /Users/...
└─┬ rimraf@3.0.2
  └─┬ glob@7.1.6
    └── inflight@1.0.6
```

Any plans for a fix?

pnappa:
I opened a fix a few days ago. #707

pnappa:
As a strawpoll, I'm curious what packages you guys are using that depend on this package? It looks like it's not been maintained for a bit, and now there's a few alternatives to use, so it may be wise to try and get those dependencies updated to use a more up-to-date package.

Personally, I use node-argon, which _used_ to depend on this, not anymore, they switched to just using `node-gyp-build`+`prebuildify`. See this commit: https://github.com/ranisalt/node-argon2/commit/b47602840a259946039db8526ddd182d1430f634#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

Luen:
Looks like [`glob`](https://github.com/isaacs/node-glob) was updated in [`rimraf`](https://github.com/isaacs/rimraf) v4 & v5 which removed the inflight package.

sabex:
This fix would be good, we use libxmljs which pulls in node-pre-gyp - so we are impacted by this vulnerability. accepting @pnappa pull request would be great.

cclauss:
% `npm audit`
# npm audit report
@babel/traverse  <7.23.2
Severity: critical
Babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code - https://github.com/advisories/GHSA-67hx-6x53-jw92
fix available via `npm audit fix`
node_modules/@babel/traverse

tar  <6.2.1
Severity: moderate
Denial of service while parsing a tar file due to lack of folders count validation - https://github.com/advisories/GHSA-f5x3-32g6-xq36
fix available via `npm audit fix`
node_modules/tar

2 vulnerabilities (1 moderate, 1 critical)

benmccann:
This should be fixed in 2.0.0-rc.0

PR #720, opened 2024-06-27
remove rimraf dependency

benmccann:
closes https://github.com/mapbox/node-pre-gyp/issues/692
closes https://github.com/mapbox/node-pre-gyp/pull/707

this removes quite a few dependencies, many of which are deprecated: https://npmgraph.js.org/?q=rimraf@5.0.5


lib/clean.js
```diff
@@ -23,7 +23,11 @@ function clean(gyp, argv, callback) {
     exists(to_delete, (found) => {
       if (found) {
         if (!gyp.opts.silent_clean) console.log('[' + package_json.name + '] Removing "%s"', to_delete);
-        return rm(to_delete, callback);
+        try {
+          fs.rmdirSync(to_delete, { recursive: true, force: true });
```
pnappa, line 27: I believe this is not correct - `rmdir` only deletes a directory if it's empty, which may not be the case here, no? https://nodejs.org/api/fs.html#fsrmdirpath-options-callback

I think it should be (also replacing the try/catch & synchronous filesystem call):
```
fs.rm(to_delete, { recursive: true, force: true }, callback);
```

lib/util/napi.js
```diff
@@ -166,16 +166,14 @@ module.exports.get_napi_build_version_from_command_args = function(command_args)
 
 module.exports.swap_build_dir_out = function(napi_build_version) {
   if (napi_build_version) {
-    const rm = require('rimraf');
-    rm.sync(module.exports.get_build_dir(napi_build_version));
+    fs.rmdirSync(module.exports.get_build_dir(napi_build_version), { recursive: true, force: true });
```
pnappa, line 169: Again, this may not work as the directory may not be empty.

lib/util/napi.js
```diff
@@ -166,16 +166,14 @@ module.exports.get_napi_build_version_from_command_args = function(command_args)
...
     fs.renameSync('build', module.exports.get_build_dir(napi_build_version));
   }
 };
 
 module.exports.swap_build_dir_in = function(napi_build_version) {
   if (napi_build_version) {
-    const rm = require('rimraf');
-    rm.sync('build');
+    fs.rmdirSync('build', { recursive: true, force: true });
```
pnappa, line 176: Also here.

pnappa (changes requested):
I prefer this PR to mine (https://github.com/mapbox/node-pre-gyp/pull/707), as removing dependencies is better than updating them.

However, I believe some code needs fixing.

pnappa:
Also, for the same reasons as in my PR, this will be a breaking change (hopefully no-one's actually affected by it), as it drops support for old versions of node. See: https://github.com/mapbox/node-pre-gyp/pull/707#issuecomment-1861812258

Minimum required version is now 14.x (not sure which version specifically)

lib/clean.js
benmccann: I think `rmdir` and `rm` are both correct, but that `rmdir` is deprecated, so I'll update them all. thanks!!

benmccann:
> Minimum required version is now 14.x (not sure which version specifically)

14.14.0

lib/util/napi.js
```diff
@@ -166,16 +166,14 @@ module.exports.get_napi_build_version_from_command_args = function(command_args)
 
 module.exports.swap_build_dir_out = function(napi_build_version) {
   if (napi_build_version) {
-    const rm = require('rimraf');
-    rm.sync(module.exports.get_build_dir(napi_build_version));
+    fs.rm(module.exports.get_build_dir(napi_build_version), { recursive: true, force: true });
```
pnappa, line 169: Apologies if you interpreted this way (I only meant to say don't use rmdirSync as opposed to rmSync, not to switch to the async version too), but by switching this to the asynchronous version the behaviour here has slightly changed. Change back to `fs.rmSync`, if it's not too much bother.

lib/util/napi.js
```diff
@@ -166,16 +166,14 @@ module.exports.get_napi_build_version_from_command_args = function(command_args)
...
     fs.renameSync('build', module.exports.get_build_dir(napi_build_version));
   }
 };
 
 module.exports.swap_build_dir_in = function(napi_build_version) {
   if (napi_build_version) {
-    const rm = require('rimraf');
-    rm.sync('build');
+    fs.rm('build', { recursive: true, force: true });
```
pnappa, line 176: Ditto, this should be `fs.rmSync`, which matches the behaviour from before.

cclauss:
Please add the modifications from #709 so we can see if the tests pass on modern versions of Node.js.
* #709

lib/clean.js
```diff
@@ -23,10 +23,11 @@ function clean(gyp, argv, callback) {
     exists(to_delete, (found) => {
       if (found) {
         if (!gyp.opts.silent_clean) console.log('[' + package_json.name + '] Removing "%s"', to_delete);
-        return rimraf(to_delete).then(
-          (result) => callback(null, result),
```
benmccann, line 27: I'm not sure why https://github.com/mapbox/node-pre-gyp/pull/707 added this line. I removed it to return the code to it's original state. @pnappa please check to make sure that I'm not overlooking something

ronilan: The [code](https://github.com/mapbox/node-pre-gyp/blame/8b99dc59ddaac4ed5158e75e70ed485aa08b07ce/lib/clean.js) prior to #707 actually errored. #707 fixed this.

In #720 (this PR) the `try .. catch` clause does the required callback.

This is 👍 

(note: test coverage needs revisit for `clean` as used in `rebuild`)

cclauss:
@ronilan @pnappa @acalcutt @benmccann Can you please help me by reviewing the open PRs on this repo?

When you think a pull request is useful and is ready to be merged, ___please consider giving it a positive review___.

If you think a pull request needs work or should not be merged, ___please consider giving it a negative review___.

Every ✅ at the top right of this page gives project maintainers confidence that the proposed changes have been read through and deemed both useful and safe to merge into the codebase.  Lots of 👍 and "_what is the ETA?_" comments are easier for maintainers to ignore than ✔️✔️✔️✔️✔️ from several different reviewers.

Similarly for ❌ on the other side.

Anyone can review a pull request on GitHub.  To do so here:
1. Scroll to the top of this page.
2. Click the `Files changed` tab and read through each file carefully looking for potential issues.
3. Click the `Review changes` button.
4. Click `Approve` (or one of the other options) and add comments only if they have not already been stated in the PR.
5. Click `Submit review` so that your ✅  or ❌ will be added to the list.

lib/clean.js
benmccann: thanks! I've added back the call to `callback`

ronilan:
> @ronilan @pnappa @acalcutt @benmccann Can you please help me by reviewing the open PRs on this repo?

@cclauss - sorry, I ran out of play time. My GitHub Sponsors page is here: https://github.com/sponsors/ronilan Good luck to all that treat legacy code maintenance a hobby rather than work.

lib/clean.js
ronilan: Early return not needed. We return at bottom when no error.

benmccann: `callback` is called with different values in each of the locations

ronilan: The [callback function](https://github.com/mapbox/node-pre-gyp/blob/master/lib/main.js#L81) receives a single argument - the error object. The [pattern](https://github.com/mapbox/node-pre-gyp/blob/master/lib/main.js#L81) used when there is no error is to call without passing a value.

pnappa (changes requested):

pnappa:
> @ronilan @pnappa @acalcutt @benmccann Can you please help me by reviewing the open PRs on this repo?

Hi Christian, I'm also not available to review PRs, I only made my PRs & reviewed the related PRs as a professional courtesy to try and patch vulnerable packages in dependencies. I no longer depend on this package, and as such can't justify helping during work hours (in my spare time I try to avoid computers). 

`node-argon2` was the package that I used that depended on `node-pre-gyp`, but they migrated to a different, more upkept package (`prebuildify`). I've talked about it in this PR before https://github.com/mapbox/node-pre-gyp/issues/705#issuecomment-1965458957 . I think community effort should be spent on consolidating onto a single product solution rather than limping along with finite resources spread too thin.

Good luck, and thanks for all the fish!

lib/clean.js
benmccann: Done. I've addressed both comments

My confusion stemmed from the use of the word async, which I usually associate with the version returning a promise rather than the version taking a callback. I think we should be sorted now though

pnappa (approved):

Alexey-Lukin (approved):

benmccann:
@cclauss this one has two approvals, so I think it should be good to merge now

cclauss:
I would rather progress on #739 before we make any removals of dependencies.  We have made a ton of changes and our test coverage is imperfect.  Let's give users a chance to kick the tires of our current state before we replace parts.
* #739

cclauss:
If you are interested in making contributions while we await progress on #739,  please consider...

Updating https://github.com/mapbox/node-pre-gyp/blob/master/README.md to reflect current versions of Node.js and our changes to automated testing.

The repo could use the addition of https://pre-commit.com or https://typicode.github.io/husky to do fast linting, type checking, spell checking, etc. at commit time.

benmccann:
> Updating https://github.com/mapbox/node-pre-gyp/blob/master/README.md to reflect current versions of Node.js and our changes to automated testing.

I've updated the supported versions of Node as part of https://github.com/mapbox/node-pre-gyp/pull/833. I'm not sure there are any change to automated testing that feel relevant to me to include in the readme. The readme never discussed automated testing in the past and readmes generally don't discuss CI setup.

> The repo could use the addition of https://pre-commit.com/ or https://typicode.github.io/husky to do fast linting, type checking, spell checking, etc. at commit time.

I would prefer not to add these. These types of checks are better on the CI in my opinion where they can be run asynchronously as opposed to interrupting the developer workflow.

> look thru the https://github.com/dependabot generated PRs that fail and add comments to those PRs about why they fail and what changes we should consider to make them pass

A lot of this effort would be wasted if done at this stage. PRs like this one remove a lot of our transitive dependencies. Rather than figuring out why an upgrade somewhere down the tree won't succeed we should first merge the PRs that remove or change dependencies. Then we can go through and figure out how to upgrade what's left.

It should be noted that the vast majority of both merged and pending dependabot PRs have no impact on users. I've sent a PR to reduce the noise from dependabot: https://github.com/mapbox/node-pre-gyp/pull/834

> Or create a changelog PR that describes the changes made for our upcoming release.

Done! https://github.com/mapbox/node-pre-gyp/pull/833

> We have made a ton of changes. Let's give users a chance to kick the tires of our current state before we replace parts.

I'm not sure I really agree with this. As you can see in the changelog PR (https://github.com/mapbox/node-pre-gyp/pull/833), we've only updated supported node versions, updated a few dependencies, and made one bug fix. This PR is less than a dozen lines and just changes the method we call to delete a file. Including it in an initial release would simplify the changelog for users because rather than there being two entries stating "updated rimraf" and "removed rimraf" in two different releases we could condense it down simply to "removed rimraf".

cclauss:
`breaking change` worries me on a codebase that is not being used... https://github.com/mapbox/node-pre-gyp/pull/720#issuecomment-2195801250  My years of effort on `node-gyp` encourages me to make small moves.

benmccann:
The breaking change was introducing the `engines` field that requires Node 18. That's already been done in an earlier PR. The APIs used in this PR only require Node 14.14.0 and so don't introduce any new breaking changes.

pnappa:
> The breaking change was introducing the `engines` field that requires Node 18. That's already been done in an earlier PR. The APIs used in this PR only require Node 14.14.0 and so don't introduce any new breaking changes.

That's right. It's a matter of philosophy whether bumping the minimum required node version is a breaking change or not - I think it is, as a user upgrading a package following semver schemes may have their app fail to compile as a result. However, we don't care about them in this scenario, as they're using EOL'd versions of NodeJS, and have much bigger problems to consider (unpatched security vulnerabilities et al).

Others disagree whether it's a breaking change or not - here, no functional run-time behaviour has changed, and as such should be fine to merge.

benmccann:
I agree that bumping the minimum node version is a breaking change, but my point is that's not being done here because we've earlier specified Node 18 as the minimum.

cclauss (approved):
Thanks for making it clear.  Let's go...

43081j:
@cclauss any chance we can get a release of this? 👀

benmccann:
The latest release was 1.1.0-dev.1, but this PR was not included in that release. Unfortunately releasing this PR by itself won't help a ton as minizlib also pulls in rimraf. That PR will need to be merged as well to see the benefit.

It's probably worth trying to get some other PRs in as well before doing a new release. @cclauss is there anything I can do to help on that front? I think several PRs in the queue are ready to go

PR #722, opened 2024-06-27
upgrade npmlog

benmccann:
the build is failing because npmlog 7 requires node 14.17, so we should first merge some of the other PRs that update the CI or I can do that here


cclauss:
Let's keep the PRs simple so that they are easy to review and merge.  Please resolve the git conflicts mentioned below.

cclauss (approved):

PR #741, opened 2024-06-29
Upgrade https-proxy-agent and nopts

benmccann:
These two have newer major versions available

There are a couple more with newer major versions available, but I think they can be removed rather than updated. Will handle that separately

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,10 +10,10 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "detect-libc": "^2.0.0",
-        "https-proxy-agent": "^5.0.0",
+        "https-proxy-agent": "^7.0.5",
         "make-dir": "^3.1.0",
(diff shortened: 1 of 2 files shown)


cclauss:
Let's have Dependabot do these upgrades.

benmccann:
Any idea why dependabot has not opened PRs for these? Do we need to update the dependabot config?

cclauss:
It wakes up once a week to generate PRs this is done to make it less chatty for repo maintainers.
* Weekly: https://github.com/mapbox/node-pre-gyp/blob/master/.github/dependabot.yml#L13
* Chatty: https://github.com/mapbox/node-pre-gyp/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduleinterval

PR #829, opened 2024-07-01
Replace `npmlog` with `consola`

br0p0p:
This PR replaces the deprecated and dependency-laden `npmlog` with a smaller, modern alternative: `consola`.

| | Module + Dependencies | Size on Disk (Unpacked Size) | Bundle size (minified) |
|--------|--------|--------|--------|
| `npmlog` | 15 | 401 KB | 65 KB |
| `consola` | 1 | 222 KB | 6 KB |
| 𝛥 | -14 (-93.3%) | -179 KB (−44.63%) | -59 KB (-90.7%) |
> This data was gathered from https://npmgraph.js.org/

Extrapolating based on the download stats for `@mapbox/node-pre-gyp` (5,730,435 downloads per week), this migration will have the following effects on the ecosystem as a whole:
- 80,226,090 fewer dependencies downloaded per week
- Reduced network traffic by ~1 TB per week

<details>
<summary><code>package-size-calculator</code> output</summary>

```
Package size report
===================

Package info for "@mapbox/node-pre-gyp@1.0.11": 2.0 MB
  Released: 2023-07-14 13:09:45.532 +0000 UTC (52w6d ago)
  Downloads last week: 4,706,924 (80.46%)
  Estimated traffic last week: 9.3 TB
  Subdependencies: 57

Removed dependencies:
  - npmlog@5.0.1: 401 kB (20.27%)
    Downloads last week: 7,502,133 (N/A% from 5.0.1)
    Downloads last week from "@mapbox/node-pre-gyp@1.0.11": 4,706,924 (N/A%)
    Traffic last week: N/A
    Traffic from "@mapbox/node-pre-gyp@1.0.11": 9.3 TB (N/A%)
    Subdependencies: 22 (38.59%)

Added dependencies:
  + consola@3.2.3: 228 kB (14.47%)
    Downloads last week: 4,630,250 (N/A% from 3.2.3)
    Estimated traffic last week: N/A
    Subdependencies: 1 (1.75%)

Estimated package size: 2.0 MB → 1.8 MB (91.26%)
Estimated traffic over a week: 9.3 TB → 8.5 TB (813 GB saved)
Estimated traffic over a week @ 100% downloads: 12 TB → 11 TB (1.0 TB saved)
Subdependencies: 57 → 36 (-21)
```

</details>

---

I will note that the output looks slightly different with these changes. Happy to update if the new format is not great

**Before**

Screenshot 2024-07-02 at 9 15 09 AM

**After**

Screenshot 2024-07-02 at 9 15 39 AM


benmccann (approved):
Amazing! Thank you!!

benmccann:
@br0p0p would you be able to fix the merge conflict?

br0p0p:
@benmccann done

lib/util/log.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+// eslint-disable-next-line node/no-missing-require
+const { createConsola } = require('consola/basic');
+
+const log = createConsola({ stdout: process.stderr });
```
benmccann, line 6: why put all output to `stderr`? that seems surprising. if we keep it we should at least add a comment to explain

br0p0p: I did this to maintain the default behavior of `npmlog`. See [here ](https://github.com/npm/npmlog/blob/756bd05d01e7e4841fba25204d6b85dfcffeba3c/lib/log.js#L12)and the [README](https://github.com/npm/npmlog/blob/756bd05d01e7e4841fba25204d6b85dfcffeba3c/README.md)

Not exactly sure why this was the default 

Probably a good call to add a comment at least. Maybe `stderr` makes sense here since the logger outputs more debugging type output and `console.log()` is used for user-facing output?

lib/install.js
```diff
@@ -149,7 +149,7 @@ function print_fallback_error(err, opts, package_json) {
     full_message += fallback_message;
     log.warn('Tried to download(' + err.statusCode + '): ' + opts.hosted_tarball);
     log.warn(full_message);
-    log.http(err.message);
+    log.log(err.message);
```
benmccann, line 152:
```suggestion
    log.error(err.message);
```

lib/util/log.js
benmccann: I think we should remove this rather than trying to maintain the behavior of `npmlog`. I checked (see below) and `consola` logs warnings and errors to stderr and info type logs to stdout, which seems like what I'd want and expect out of a logging system. Let's just check that we're setting the log levels appropriately whenever we do logging - I left one comment where I think we should do `log.error` to make sure an error goes to stdout

https://github.com/unjs/consola/blob/713ec3454889bf967e9c694bf800955f1cd60e52/src/reporters/basic.ts#L70
https://github.com/unjs/consola/blob/713ec3454889bf967e9c694bf800955f1cd60e52/src/constants.ts#L23

br0p0p: Sure thing. Updated.

My thinking was to prevent potentially breaking dependents which rely on the `stderr` output for whatever reason but I agree that not using `stderr` for everything makes way more sense

benmccann (approved):
lgtm except that it looks like the appveyor tests might be failing

@cclauss could you kick off the github actions tests on this PR?

benmccann:
@br0p0p I'm afraid all the CI jobs are failing with this PR. would you be able to take a look?

benmccann (dismissed):
Thank you @br0p0p !

@cclauss this one is now ready to go. It would be great to merge before the next preview release as it fixes a few deprecation warnings (I believe all the remaining).

lib/util/log.js
lukekarrys: The main reason that most logging from `npm` and `node-gyp` goes to stderr (even now that neither use `npmlog`) is that stdout is reserved for output that is expected to be used programmatically (usually json). We don't usually swap which output stream is used based on flags like `--json` so the default is to place all logging on stderr so that if a json command is piped to something like `jq` the logging won't inadvertently create invalid json.

There are a lot of gray areas to this rule and this is something I spent a lot of time figuring out for `npm` in the past. I haven't actually review this code (yet) so I'm not saying that all output should be put on stderr, just that as a default stderr is usually better than stdout for CLI logging output.

br0p0p: Thanks for this context @lukekarrys. Thoughts @benmccann?

benmccann: I have no objection to `stderr` given that additional context. If @lukekarrys thinks it's the way to go I'll defer to him on that

benmccann: @br0p0p do you want to change this back to use `stderr` by default? It might be the safer option to avoid changing that

br0p0p: @benmccann sure, no problem

HummingMind:
Hoping this gets merged soon. Thank you for your work! 🍻

lib/util/log.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const { createConsola } = require('consola/basic');
+
+const log = createConsola({ stdout: process.stderr });
```
benmccann, line 5:
```suggestion
// match the default behavior of npm and node-gyp  where stdout is reserved for output that is expected
// to be used programmatically (usually json)
const log = createConsola({ stdout: process.stderr });
```

benmccann (approved):

benmccann (approved):

PR #833, opened 2024-07-03
update changelog for pending release

benmccann:

Issue #739, opened 2024-06-29
Needed: A GitHub Action to package and publish a new version


cclauss:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs#packaging-workflow-data-as-artifacts

-->
*  #744

acalcutt:
In maplibre-native workflow we settled on this workflow that checks if the version in package.json is published. 
https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-release.yml#L9-L34

If it is not published it runs the build
https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-release.yml#L38

maybe something like that would work here. Seems like it would be easier because this project doesn't really need to build anything, just run the publish steps.
https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-release.yml#L304-L310

cclauss:
Cool!  Pull requests will be reviewed.  Let's see how that approach works.

mapsam:
I'm in favor of GitHub release for this as well. This may be blocked by using the `@mapbox/node-pre-gyp` namespace for a bit, since the Mapbox NPM scope is pretty locked down and I'm not sure we'll be able to grant tokens for public repositories like this. Ultimately this is a good argument to move back to the unscoped `node-pre-gyp` NPM module which does not have the same security restrictions.

Not ideal, but in the short-term I can cut one-off prereleases for `@mapbox/node-pre-gyp` as needed @cclauss

cclauss:
I understand the complexities.  I think we need a pre-release and then a release BEFORE we merge other changes (all the Dependabot changes, etc.).  Please proceed with that pre-release when you feel comfortable doing so.

mapsam:
Sounds good! Can you give me a commit to prerelease from? Just want to make sure I release what you're expecting. Latest master is https://github.com/mapbox/node-pre-gyp/commit/13d4ac74853070cd2a9649e1f8079e5d68839c7a

mapsam:
`@mapbox/node-pre-gyp@1.1.0-dev.1` has been published from https://github.com/mapbox/node-pre-gyp/commit/13d4ac74853070cd2a9649e1f8079e5d68839c7a.

I need to push package.json version changes to GitHub in order to publish to the `@mapbox/` namespace, so technically the version is published from https://github.com/mapbox/node-pre-gyp/commit/996b31b06800d384e893c1108b7e8796cb97cd8f. 

[...]

PR #688, opened 2023-06-28
Add Dependabot configuration file to keep dependencies up to date

striezel:
Add Dependabot configuration to automatically open PRs for outdated NPM packages.

Currently, a lot of dependencies are outdated, some even contain security vulnerabilities (according to `npm audit`), so this might help to improve the situation.

PR #834, opened 2024-07-03
Dependabot should only open PRs for meaningful updates

benmccann:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow:

> By default all dependencies that are explicitly defined in a manifest are kept up to date by Dependabot version updates. In addition, Dependabot security updates also update vulnerable dependencies that are defined in lock files. 

This would be much better than our current config. Every single dependabot PR that's passing the CI currently looks to be just noise as none of them affect users and only affect our own lockfile. This makes it difficult to tell which PRs are actually meaningful as they're hidden amongst the ones with no impact.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,12 @@ version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/"
+    groups:
+      github-actions:
+        patterns:
+          - "*"  # Group all Actions updates into a single larger pull request
     schedule:
       interval: "daily"
-    # Allow up to 10 open pull requests for dependencies.
-    open-pull-requests-limit: 100  # TODO: Revert to 10
   - package-ecosystem: github-actions
     directory: /
     groups:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,9 +7,6 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-    allow:
-      # Allow both direct and indirect updates for all packages.
-      - dependency-type: "all"
     # Allow up to 10 open pull requests for dependencies.
     open-pull-requests-limit: 100  # TODO: Revert to 10
   - package-ecosystem: github-actions


cclauss:
Agreed.  I did this on purpose temporarily to see where we had problem dependencies and where we did not.

Once we have a v2 production release and then our dependencies upgrades have settled down, I would prefer that JS dependencies are grouped like our GitHub Actions are currently configured.

striezel:
Since I am the one who opened the pull request that introduced the Dependabot configuration (https://github.com/mapbox/node-pre-gyp/pull/688), I may at least in part be to blame for the current situation. However, in my pull request the limit was set so that Dependabot only opens at most ten pull requests at a time. In my experience that limit is a good way to make sure maintainers do not get overwhelmed by the sheer amount of incoming pull requests from Dependabot. Then somebody else increased the limit to 100, and as a result there are currently over 70 open Dependabot pull requests.

That is a huge number, but something like that can unfortunately be expected. Until a week ago the latest commit on the `master` branch was from 14 July 2023, so there is almost a full year of version updates to catch up with. In a fast-moving ecosystem like npm / JavaScript that can be a lot. Basically, there was a backlog of updates building up for almost a year. It's just that Dependabot now makes that visible to us. But as soon as those updates have been dealt with, the number of new incoming Dependabot updates should be relatively low, even with the current configuration.

One thing that helps in dealing with so much version updates is a comprehensive and automated test suite.
So the question is: Do you trust your test suite to catch possible problems that may arise when upgrading dependencies to newer versions? If the answer to that question is yes, then you can basically merge any Dependabot pull request that passes all tests. Seeing that most of the Dependabot pull requests passed, one could probably merge 60 or more of those 76 pull requests now, assuming that the test suite is good.

benmccann:
The bigger problem is that none of the updates are meaningful regardless of how gradually you deal with them. See https://github.com/mapbox/node-pre-gyp/pull/833

cclauss:
> somebody else increased the limit to 100

Guilty.

> Do you trust your test suite to catch possible problems

I have less confidence than others.

> One could probably merge 60 or more of those 76 pull requests now

Let's do that just after #739 is completely resolved and we are on a new production release.  After that megamerge, we can group the JS updates just like #740.

> None of [the open Dependabot PR] will affect users.

How do we prove that?

Thanks for your help on this.

benmccann:
> How do we prove that?

Because they don't edit `package.json`, which is what gets shipped to users. They only update `package-lock.json`, which isn't part of the distributable and is only used for our own testing

benmccann:
Can we please merge this? Dependabot is clogging up the PR queue again

PR #847 and everything after that should be closed as none of them have any meaningful effect. If we merge them they will clog the commit log with noise. Some of the earlier dependabot PRs may be nice to merge eventhough most of those also don't affect users. However, they will help keep our testing infrastructure updated by bumping packages to new major versions. I wouldn't object if those are closed, but feel they can potentially be worth it as doing major version upgrades can help keep things modernized. However, #847 and after are all minor/patch versions and should really be avoided because we simply can't have 4-5 new PRs opened every day

cclauss (approved):

PR #836, opened 2024-07-03
update lockfile

benmccann:
update from lockfile version 2 to version 3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@mapbox/node-pre-gyp",
   "version": "1.0.11",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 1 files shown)


striezel:
It's probably a good idea to wait for all (or at least most) of those open dependabot pull requests to get merged before these changes get merged. Otherwise dependabot might rebase its open pull requests, and that will trigger a lot of CI re-runs. That's probably okay for builds on GitHub Actions, but the AppVeyor builds usually take a while, at least in my experience.

benmccann:
I think the more likely outcome is that it would close the majority of the pending PRs as being no longer necessary since those PRs only affect the lockfile and this PR updates the lockfile. I think it's better to handle them in one go like this or like rennovate bot does than to do 50 individual PRs that just clog up the PR queue and commit log

cclauss (approved):

PR #904, opened 2025-05-07
Update abi_crosswalk.json for abi 137 / node 24

acalcutt:
This PR updates the abi_crosswalk.json in the repo to support node 24 / ABI 137 

This technically gets updated as part of a release, but I think it it good to update it in the repo as well.


acalcutt:
I can make a PR for a release soon so we can get one with this included if you would like, maybe like a 2.0.1 release

The only thing I meant to ask about was the lts/* in the ci. Right now that seems to resolve to node 22, so I added 24.x in the list, but i would imagine eventually that will be lts/* also. it might be better to just have 20.x, 22.x, and 24.x

cclauss:
A new release makes sense to me.
* nodejs/nan/#997

The reason that I would like to keep lts/* is because this repo has few maintainers and has sat unmaintained for long periods of time, and it would be better to have proof of broken new LTS versions than have them be silently broken.

PR #908, opened 2025-06-17
fix moduleName environment for multiple dashes

### LINKED ISSUE #896: binary host mirror override does not work for module_names with multiple dashes
The binary host mirror environment variable does not work for packages with `binary.module_name` values containing more than one dash because it does not use `replaceAll` here: https://github.com/mapbox/node-pre-gyp/blob/master/lib/util/versioning.js#L315. I encountered this trying to use the binary host mirror override with https://github.com/confluentinc/confluent-kafka-javascript/blob/master/package.json#L19.

baswag:
Fix #896 by using `replaceAll` instead of `replace`


cclauss:
> Is there anything I can do to help?

Please consider creating a ___SEPARATE___ PR that gets the failing tests below (npm audit?) to pass.

Issue #924, opened 2025-11-26
v2.0.2 introduced a breaking change in a patch: now requires `node@^20.17.0 || >=22.9.0`


serhalp:
v2.0.2 was published a few hours ago as a patch, but it includes [this bump of `nopt` from `^8.0.0` to `^9.0.0`](https://github.com/mapbox/node-pre-gyp/commit/c842373a353e7c9478152bd9e36f6de02d59f521), which in turn [bumped the supported node.js version range from `^18.17.0 || >=20.5.0` to `^20.17.0 || >=22.9.0`](https://github.com/mapbox/node-pre-gyp/commit/c842373a353e7c9478152bd9e36f6de02d59f521).

This is a breaking change, but `@mapbox/node-pre-gyp` did not receive a major bump.

It might be too late to unpublish this release at this point. Could you please revert this change in a follow-up patch?

This is breaking downstream consumers of this very popular package.

Thanks!

PR #925, opened 2025-11-26
fix: revert "Bump nopt from 8.0.0 to 9.0.0 (#919)"

serhalp:
This reverts commit c842373a353e7c9478152bd9e36f6de02d59f521.

Fixes https://github.com/mapbox/node-pre-gyp/issues/924.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "detect-libc": "^2.0.0",
         "https-proxy-agent": "^7.0.5",
         "node-fetch": "^2.6.7",
-        "nopt": "^9.0.0",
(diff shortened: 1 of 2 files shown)


rafaykh90:
@benmccann I cut the last release. My change in https://github.com/mapbox/node-pre-gyp/pull/923 was not supposed to be a breaking change therefore I didn't make a major version change. The breaking change was introduced in one of the earliers PRs which was merged. 

As for the change in this PR, do you think the version should be bumped to 2.0.3? I am not sure if it is even allowed on the registry to unpublish a package.

PR #929, opened 2025-12-12
Migrate NPM publishing to use trusted publishing instead of classic tokens

### LINKED ISSUE #928: Move node release workflow to use trusted publishing
The tokens in the node workflow won't work anymore because npm has sunset the classic tokens. 

https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/

like used in https://github.com/mapbox/node-pre-gyp/blob/master/.github/workflows/release.yml#L91-L105
See migrations like:
https://github.com/maptiler/tileserver-gl/pull/1872

acalcutt:
NPM has sunset classic tokens as of December 9, 2025. This updates the release workflow to use OIDC-based trusted publishing instead. This could also help with our issue keeping the token up to date.

### Workflow permissions
- Added `id-token: write` permission for OIDC token generation
- Moved `contents: write` to top-level permissions

### NPM authentication
- Removed manual token management (`npm config set` and `NPM_TOKEN` env vars)
- Added `registry-url: 'https://registry.npmjs.org'` to `setup-node` action
- Simplified publish commands to use automatic OIDC authentication

### Package configuration
- Updated `package.json` repository URL from `git://` to HTTPS format (required for trusted publishing)

## Prerequisites for deployment

Before this workflow can be used, the NPM package must be configured for trusted publishing:

1. Go to https://www.npmjs.com/package/@mapbox/node-pre-gyp/access
2. Navigate to the "Publishing" section
3. Configure trusted publishing with GitHub Actions
4. Add `mapbox/node-pre-gyp` repository and `release.yml` workflow as trusted sources

The workflow will then automatically receive and use OIDC tokens from GitHub Actions.

## Benefits

- **Enhanced security**: No more storing long-lived tokens as secrets
- **Automatic token management**: GitHub Actions generates short-lived tokens automatically
- **Compliance**: Aligns with npm's new authentication requirements

## References

- NPM announcement: https://github.blog/changelog/2025-12-09-npm-classic-tokens-revoked-session-based-auth-and-cli-token-management-now-available/
- Similar migration: https://github.com/maptiler/tileserver-gl/pull/1872

Fixes #928

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,9 +41,6 @@ jobs:
     needs: release-check
     if: ${{ needs.release-check.outputs.published == 'false' }}
     runs-on: ubuntu-latest
-    permissions:
-      contents: write # allow github publish action
-      id-token: write # Enables OIDC token generation
     defaults:
       run:
         shell: bash


.github/workflows/release.yml
```diff
@@ -48,6 +50,7 @@ jobs:
       - uses: actions/setup-node@v6
         with:
           node-version: 22
```
lukekarrys: Node 22 ships with npm 10.9.4. Ref: https://nodejs.org/en/about/previous-releases But Trusted Publishing requires npm 11.5.1 or higher.

So I think this workflow would also require updating npm with `npm install npm@11 -g`.

acalcutt: second issue i mentioned is fixed. i had accidently remvoed contents: write from the publish step so that is what the other action was failing

https://github.com/acalcutt/node-pre-gyp/actions/runs/20175726794/job/57923175131

lukekarrys: Looks good! Updating to Node 24 will get us npm 11.6.2 which should work.

acalcutt:
Maybe @rafaykh90 could help set up the since they had the permissions to publish. Someone with the rights needs to add the trusted publisher in npm settings. On my test package I set it up like this

image

I think here we could also limit it to the main environment

.github/workflows/release.yml
```diff
@@ -1,5 +1,9 @@
 name: release
 
+permissions:
+  id-token: write  # Required for NPM trusted publishing
+  contents: write
```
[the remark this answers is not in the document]
acalcutt: remvoed that block and it works fine

acalcutt:
If we want to test the existing release workflow with the updated npm key, the version would need to be bumped. it only publishes when the version does not exist on npm. Basically following https://github.com/mapbox/node-pre-gyp/blob/master/RELEASE.md

I could make a PR to bump to a prerelease to test if you wanted

rafaykh90:
@acalcutt Feel free to do so. I just want to make sure that the CI is fixed for now and we are unblocked.

PR #930, opened 2025-12-12
pin @mapbox/eslint-config-mapbox to 5.0.1

lukekarrys:
Later versions depend on `@mapbox/eslint-plugin-mapbox` which doesn't appear to be in the npm registry.

This dep was already effectively pinned because of the lockfile, but when I attempted to install without a lockfile from my local machine it errored.


cclauss:
git conflicts

benmccann:
We should probably also file an issue letting them know about the problem

PR #945, opened 2026-03-11
Migrate off deprecated `url.parse` APIS to WHATWG URL API instead.

arthurfiorette:
This PR attempts to remove the following warnings when using this library:

```
(node:960561) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

Node itself provides a codemod (`npx codemod@latest @nodejs/node-url-to-whatwg-url`) that mainly replaces `url.parse`.

`url.resolve` is also widely used and their d[ocs provide an example](https://nodejs.org/api/url.html#urlresolvefrom-to) on how to replace it using the modern WHATWG URL API.

### DIFF
--- a/lib/install.js
+++ b/lib/install.js
@@ -11,7 +11,7 @@ const existsAsync = fs.exists || path.exists;
 const versioning = require('./util/versioning.js');
 const napi = require('./util/napi.js');
 const s3_setup = require('./util/s3_setup.js');
-const url = require('url');
+const url = require('./util/url.js');
 // for fetching binaries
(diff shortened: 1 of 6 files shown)


arthurfiorette:
Fixed lint issues

arthurfiorette:
Fixed test path url issues

PR #953, opened 2026-04-07
Avoid npm warnings by supporting MODULE_BINARY_HOST_MIRROR env variable

yunnysunny:
## Description
With newer versions of npm (v10+), using environment variables in the form of:
```
npm_config_<module>_binary_host_mirror
```
will trigger warnings such as:
```
npm warn Unknown env config "<module>-binary-host-mirror"
```
This happens because npm now validates environment variables prefixed with npm_config_* against its known configuration keys, and custom keys used by tools like node-pre-gyp are no longer recognized.

## Changes

To address this issue, this PR introduces support for an alternative environment variable:
```
<MODULE>_BINARY_HOST_MIRROR
```
Specifically:

The code now checks for `npm_config_<module>_binary_host_mirror` 
Falls back to `process.env.<MODULE>_BINARY_HOST_MIRROR`


yunnysunny:
It's been some time since this PR was proposed. Could we consider this pull request now? @cclauss

PR #963, opened 2026-05-25
Remove node-fetch dependency and use built-in https/http modules instead

Shooteger:
This package here already requires node >= 18, so node-fetch is just dead weight so I replaced it with the built-in https/http modules in lib/install.js

Also fixed some bugs while doing this:
- ca and cafile were read from opts but never actually passed to fetch() which was silently ignored before this PR
- same for the User-Agent header, built but thrown away
- follow_max: 10 in requestOpts was dead code, node-fetch never understood it,
  so redirect loops would just hang forever (its 20 I think, in node-fetch v2)

```js
// ca is written into requestOpts...
// follow_max is set in requestOpts...
const requestOpts = {
  uri: sanitized,
  headers: { 'User-Agent': '...' },
  follow_max: 10  // node-fetch has never had this option
};

// ...but fetch() only receives { agent }, so follow_max is thrown away too.
// node-fetch's actual option is `follow`, not `follow_max`.
// Without it, node-fetch defaults to 20 redirects — not 10 as intended.
fetch(sanitized, { agent })
```

The new fetch_with_redirects() helper fixes all of that. It brings proper 10 redirect cap,
ca/agent/headers all wired through correctly, returns an IncomingMessage so
.pipe() into tar works directly.

native fetch in Node 18+ uses undici under the hood, not the http/https modules.
nock only intercepts http/https, so the whole test suite would break silently. 

Also ran `node scripts/abi_crosswalk.js` and committed the changes

930 tests (all) passing.

Please review and let me know, if it can be merged

MP


Shooteger:
For reference, in src/index.js for node-fetch 2.x, we have this:

```
// HTTP-redirect fetch step 6 (counter increment)
// Create a new Request object.
const requestOpts = {
    headers: new Headers(request.headers),
    follow: request.follow,  // <-- the option is called `follow`, not `follow_max`
    counter: request.counter + 1,
    agent: request.agent,
    ...
};
```
with counter
```
if (request.counter >= request.follow) {
    reject(new FetchError(`maximum redirect reached at: ${request.url}`, 'max-redirect'));
```

PR #999, opened 2026-09-17
chore: replace the end-of-life Node 20 in the CI test matrices

X-Guardian:
Per [@cclauss's request](https://github.com/mapbox/node-pre-gyp/issues/998#issuecomment-5701508006), this replaces the end-of-life Node 20 with Node 22 in the CI test matrices.

Node 20 reached end-of-life on 2026-04-30, Node 22 is in maintenance until 2027-04-30

PR #1001, opened 2026-09-17
Fix Windows CI against Visual Studio 2026 and Node 26

### LINKED ISSUE #998: All Windows CI jobs fail: due to `windows-latest` now shipping Visual Studio 2026
## Summary

Every `windows-latest` job in `ci.yml` fails at `configure`. The GitHub Actions `windows-latest` image now serves Visual Studio 2026 (18.x) and no longer carries Visual Studio 2022.

There are two independent causes, split across the matrix:

1. **Node 20.x** — the bundled node-gyp (10.1.0) predates VS 18 support and cannot detect the installed compiler at all.
2. **Node 24.x and `lts/*`** — the bundled node-gyp (12.4.0) detects VS 2026 correctly, but [`test/run.util.js:63`](../test/run.util.js#L63) hardcodes `--msvs_version=2022`, so the detected toolchain is rejected.

## Evidence

Failing run: https://github.com/mapbox/node-pre-gyp/actions/runs/35073143327 (all three Windows jobs fail)

### The image no longer has VS 2022

`windows-latest` resolves to the `win25-vs2026` image:

- Operating System: Microsoft Windows Server 2025
- Image Release: https://github.com/actions/runner-images/releases/tag/win25-vs2026%2F20260907.229

### Cause 1 — Node 20.x: node-gyp cannot see VS 18

```
gyp ERR! find VS unknown version "undefined" found at "C:\Program Files\Microsoft Visual Studio\18\Enterprise"
gyp ERR! find VS could not find a version of Visual Studio 2017 or newer to use
```

node-gyp gained VS 18 support in **12.0.0**. In `lib/find-visualstudio.js`, 12.4.0 has:

```js
if (ret.versionMajor === 18) {
  ret.versionYear = 2026
  return ret
}
```

This block is absent in 10.1.0 and 11.4.2, where the version ladder stops at `versionMajor === 17`. Verified by diffing `find-visualstudio.js` across 10.1.0, 11.4.2, 12.4.0 and 13.0.2.

Which node-gyp each job gets, via the bundled npm:

| Matrix entry | Node | npm | node-gyp | VS 18 detected? |
|---|---|---|---|---|
| `lts/-1` | 22.23.2 | 10.9.8 | 11.5.0 | No |
| `lts/*` | 24.21.0 | 11.19.0 | 12.4.0 | Yes |
| `latest` | 26.9.0 | 11.19.1 | 12.4.0 | Yes |

Only `lts/-1` (Node 22) is affected. These aliases float: `lts/*` moves to Node 26 when it enters LTS in October 2026, at which point `lts/-1` becomes Node 24 and the problem disappears on its own — but Node 22 is supported until 2027-04-30, so it needs handling until then.

### Cause 2 — Node 24.x / `lts/*`: hardcoded `--msvs_version=2022`

node-gyp 12.4.0 finds the toolchain without trouble, then rejects it only because of the pinned version flag:

```
gyp ERR! find VS checking VS2026 (18.9.12120.119) found at:
gyp ERR! find VS - found "Visual Studio C++ core features"
gyp ERR! find VS - found VC++ toolset: v145
gyp ERR! find VS - found Windows SDK: 10.0.26100.0
gyp ERR! find VS - "2026"
gyp ERR! find VS - looking for Visual Studio version 2022
gyp ERR! find VS - msvs_version does not match this version
gyp ERR! find VS could not find a version of Visual Studio 2017 or newer to use
```

The flag comes from [`test/run.util.js:63`](../test/run.util.js#L63):

```js
// Test building with msvs 2022
if (process.platform === 'win32') {
  final_cmd += ' --msvs_version=2022 ';
}
```

Consistent with there being only one blocker on Node 24, it fails fewer tests (`not ok 4`) than Node 20 (`not ok 2`, `4`, `7`, `8`, …), which fails earlier and more broadly.

## Possible directions

**The question to settle first: should node-pre-gyp support building against VS 2026 generally, or is this only about making CI green?** Every option below is downstream of that.

Relevant to it: `engines` is `>=18`, and the README states support tracks the [Node.js release schedule](https://github.com/nodejs/release#release-schedule), retiring versions at EOL. **Node 20 went EOL on 2026-04-30**, so the matrix is currently testing an EOL release — which bears directly on how much the Node 20 breakage is worth working around.

- **Remove the `--msvs_version=2022` flag** from `test/run.util.js`, letting node-gyp auto-detect. Plausibly fixes Node 24 and `lts/*`. Does nothing for Node 20, which fails before the version check.

  The flag looks incidental rather than deliberate. It was introduced as `--msvs_version=2015` with the comment "since that is more edge case than 2013" — the intent was to exercise a non-default VS version, and the current `2022` value arrived via a mechanical bump in #707. Hardcoding a version means the value needs updating every time the runner image moves, which is how this broke. Auto-detection removes that maintenance entirely.

  Trade-off: `test/run.util.js:63` is the only reference to `msvs_version` under `test/`, so removing it drops the sole coverage of node-pre-gyp's `--msvs_version` pass-through ([`lib/configure.js:15`](../lib/configure.js#L15) lists it among `known_gyp_args`). If that path is worth testing, it would be better covered by an explicit unit test of argument forwarding than by a global flag on every Windows build.
- **Force a newer node-gyp for Node 20** (e.g. `npm_config_node_gyp` pointing at node-gyp ≥ 12). Would fix Node 20, but means CI no longer exercises the node-gyp that Node 20 users actually get — worth weighing, since that is part of what this package integrates with.
- **Drop Node 20 from the matrix**, since it is already past EOL and the README commits to retiring EOL versions. This removes Cause 1 entirely rather than working around it, leaving only the `--msvs_version=2022` flag to fix. Note this is a support-policy change, not just a CI change — `engines` is currently `>=18`.

X-Guardian:
Closes #998, which has the diagnosis: `windows-latest` now ships Visual Studio 2026, the tests hardcode `--msvs_version=2022`, and node-gyp only recognises VS 18 from v12 onwards.

### Changes

**Drop the hardcoded `--msvs_version=2022`** from `test/run.util.js` and let node-gyp detect the toolchain. Each runner image carries exactly one Visual Studio, so there is nothing to disambiguate: the flag only ever asserted a year that happened to match. It was not deliberate either. It began as `--msvs_version=2015` with the comment "since that is more edge case than 2013", and the `2022` value arrived via a mechanical bump in #707, needing an edit on every image change since.

**Add `windows-2022` to both matrices.** VS 2022 remains widely used and [is the image GitHub directs users to](https://github.com/actions/runner-images/issues/14017) when they need it, so it is worth covering alongside VS 2026 rather than testing only the newest toolchain.

**Add `scripts/upgrade-node-gyp.js`,** run on Windows via `npm run upgrade-node-gyp`. It reads the version of the node-gyp that node-pre-gyp would invoke and installs a newer one only when that version predates VS 2026 support. Keying off the resolved node-gyp rather than the Node version matters because the matrix uses floating aliases: `lts/-1` is Node 22 today and becomes Node 24 when Node 26 enters LTS, at which point the script finds node-gyp 12 and does nothing.

**Skip one test where npm's own node-gyp is too old.** npm runs install hooks against the copy it bundles, ignoring `npm_config_node_gyp` whether set by environment variable, `.npmrc`, or `--node-gyp`. Where that copy cannot detect the installed Visual Studio, `app1 passes --nodedir down to node-gyp via npm` fails at VS detection before `--nodedir` is read, so the assertion cannot be made. The sibling test covers the same forwarding via node-pre-gyp directly.

**Regenerate `abi_crosswalk.json`** through Node 26.10.0. The `latest` matrix entry now resolves to Node 26, which failed with `Unsupported target version: 26.9.0` on every platform.

### node-gyp resolution in the tests

npm resets `npm_config_node_gyp` to its bundled copy whenever it runs a script, so a value exported by the upgrade step does not survive `npm test`. The script therefore exports `NPG_TEST_NODE_GYP`, a name npm leaves alone, and `test/run.util.js` passes it to each child as `npm_config_node_gyp`, which node-pre-gyp already reads.

This keeps the change out of the published package: `lib/util/compile.js` gains one line exporting `which_node_gyp`, so the upgrade script resolves node-gyp exactly as the build does. No resolution behaviour changes for consumers.

### Result

15 jobs in `ci.yml` and 12 in `s3-bucket.yml`, with Windows covered on both toolchains:

| Node | Currently | windows-latest (VS 2026) | windows-2022 (VS 2022) |
|---|---|---|---|
| `lts/-1` | 22.23.2 | node-gyp upgraded | bundled node-gyp |
| `lts/*` | 24.21.0 | bundled node-gyp | bundled node-gyp |
| `latest` | 26.9.0 | bundled node-gyp | bundled node-gyp |

Non-Windows jobs are unaffected. The upgrade step is guarded by `runner.os == 'Windows'`, and `--msvs_version` was only ever passed under a `win32` check.

### Notes for reviewers

- The node-gyp version installed by the upgrade is pinned in the script as `INSTALL_VERSION` rather than tracking `@latest`, so CI does not silently follow new node-gyp releases.
- The install is local and `--no-save`. A global install is not picked up: `which_node_gyp` resolves through `npm_config_node_gyp`, then `require.resolve`, then npm's bundled copy, none of which see the global prefix. Verified by installing node-gyp 13.0.2 globally and confirming resolution still returned npm's bundled copy.
- `s3-bucket.yml` also picks up a fix for `${{ matrix.node }}` references left behind by the matrix key rename in #999. Its tests are skipped when `S3_BUCKET` is unset, so the Windows changes there are untested in forks.
- `npm run update-crosswalk` remains commented out in `ci.yml`, so the crosswalk will go stale again on the next Node release.


cclauss:
# Are you ready for me to merge this?

X-Guardian:
All good!

lib/util/abi_crosswalk.json
```diff
@@ -3299,8 +3395,240 @@
...
+  "26.9.0": {
+    "node_abi": 147,
+    "v8": "14.6"
+  },
+  "26.10.0": {
+    "node_abi": 147,
+    "v8": "14.6"
   }
-}
+}
```
cclauss:
```suggestion
}

```

cclauss (dismissed):

cclauss (approved):

